PR #199, opened 2015-01-05
[discuss] TIFF output for raster tables

mojodna:
This adds GeoTIFF as an output format.

This also makes sense as a Windshaft output format, but seems fairly sensible here.

### DIFF
--- a/app/models/formats/tiff.js
+++ b/app/models/formats/tiff.js
@@ -1,0 +1,29 @@
+"use strict";
+
+var util = require("util");
+
+var pg = require("./pg");
+
+var TIFF = function() {};
+
+TIFF.prototype = new pg("tiff");
+
+TIFF.prototype._contentType = "image/tiff";
+
+TIFF.prototype.getQuery = function(sql, options) {
+  // the query will contain conversion from coords to a bbox
+  // NOTE: gn is assumed to be the raster column rather than the geometry
+  // column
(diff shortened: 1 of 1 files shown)


app/models/formats/tiff.js
```diff
@@ -0,0 +1,29 @@
...
+
+TIFF.prototype = new pg("tiff");
+
+TIFF.prototype._contentType = "image/tiff";
+
+TIFF.prototype.getQuery = function(sql, options) {
+  // the query will contain conversion from coords to a bbox
+  // NOTE: gn is assumed to be the raster column rather than the geometry
+  // column
+  return util.format("SELECT ST_AsTiff(%d, '') tiff FROM (%s) AS _ LIMIT 1", options.gn, sql);
```
javisantana: For others format renderers that requires geometry we use ``the_geom`` (default value for ``gn``) and if the table of query does not contain that column it justf fails (http 400)

In this case ``gn`` should be ``the_raster`` instead of ``the_geom``

This question opens a new more general one: how sql api should manage raster tables?

javisantana:
for me this is ok except for the_raster thing explained in the comment.

Other thing that changes is that other columns from the query are discarded, but that could be explained in the documentation.

Did you tried it?

mojodna:
No, that's the other thing ;-)  I still need to set up a working dev environment.

Algunenano:
If necessary it should be implemented in Windshaft

PR #200, opened 2015-01-06
[discuss] Mapbox vector tile (PBF) output

mojodna:
This is both incomplete and not really a good idea, but it brings up some points worth discussing, I think.

While writing this (and the TIFF patch), I started wondering about the distinction between the SQL API and the Windshaft-powered endpoints.  Specifically, there feels like there's a lot of overlap between the 2.

The Windshaft endpoints seem like variants of the SQL endpoints that produce tiled data (and images, but I can't think of a good reason that the SQL endpoint shouldn't include those).

This is partially to say that this output format is better suited to Windshaft (but that wasn't where I initially thought to implement it), since it requires that query parameters be passed through (in order to render the PBF) **and** coordinates need to be embedded into the SQL query that's provided (which would be implicit if powered by Windshaft).

Have you guys given any thought to merging the functionality of the 2 sets of endpoints?

### DIFF
--- a/app/controllers/app.js
+++ b/app/controllers/app.js
@@ -467,7 +467,8 @@ function handleQuery(req, res) {
                   filename: filename,
                   bufferedRows: global.settings.bufferedRows,
                   callback: params.callback,
-                  abortChecker: checkAborted
+                  abortChecker: checkAborted,
+                  params: params
                 };
 
                 if ( req.profiler ) {
--- a/app/models/formats/pbf.js
+++ b/app/models/formats/pbf.js
@@ -1,0 +1,105 @@
+"use strict";
+
+var fs = require("fs"),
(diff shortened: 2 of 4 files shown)


javisantana:
If you think about any API it can be probably done with the SQL API, at the end is a wrapper of postgres so it's tempting to add features and output formats in the SQL API.

The general rule here is:

- everything that needs some process to render on a map -> windshaft
- general data (row oriented) -> SQL API

For this case I think is better to use the incoming postgis functions to export MVT: http://trac.osgeo.org/postgis/ticket/2910

In general windshaft is aware of specific stuff: caching, security, layer merging, fetching external resources (lot of things are coming), take care of the resources used to render and so on.

Algunenano:
Windshaft has MVT support

PR #205, opened 2015-01-30
added /u/:user routing

javisantana:


javisantana:
cc @rochoa

app/controllers/app.js
```diff
@@ -164,13 +164,24 @@ app.set("trust proxy", true);
...
-  res.send(getVersion());
-});
+
+app.all(global.settings.base_url + '/sql',  handleQuery);
+app.all(global.settings.base_url + '/sql.:f',  handleQuery);
+app.get(global.settings.base_url + '/cachestatus', handleCacheStatus);
+app.get(global.settings.base_url + '/health',  handleHealthCheck);
+app.get(global.settings.base_url + '/version', handleVersion);
+
+if (global.settings.user_url) {
```
rochoa: I think we can avoid this by making /u/:user as optional in the route path

```javascript
base_url = '/(u/:user)?/api/:version'
```

app/controllers/app.js
```diff
@@ -207,6 +222,8 @@ function handleQuery(req, res) {
     var requestedFilename = params.filename;
     var filename  = requestedFilename;
     var requestedSkipfields = params.skipfields;
+    // if the request contains the user use it, if not guess from the host
+    var cdbUsername = req.params.user || cdbReq.userByReq(req);
```
rochoa: I would keep the user extraction inside userByReq.

Off topic: userByReq should live in a common module for CartoDB-SQL-API and Windshaft-cartodb.

rochoa:
I made some changes with my proposal, let me know what you think about them :smile:

javisantana:
:es:

PR #207, opened 2015-02-13
Feature/remove cartodb auth

flibbertigibbet:
* Disable CartoDB user authentication via redis and subdomain (rely on database permissions instead)
* Switch geo column name from `the_geom` to `geom` 
* Document additional database functions required from [this repo](https://github.com/CartoDB/cartodb-postgresql)

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,6 +28,14 @@ Install dependencies
 npm install
 ```
 
+Add CartoDB functions to PostgreSQL database
+--------------------------------------------
+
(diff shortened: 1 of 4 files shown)


flibbertigibbet:
Sorry!  Meant to open this PR against a fork.

javisantana:
It would be better to implement a interface for metadatadb so you can implement it using a ConfigBasedMedataDB, right?

(I know this is not for our fork but it would be nice to have this on the official fork)

PR #216, opened 2015-04-06
Disable per user health check

lbosque:


rochoa:
I'm fine as far as you remove the commented out code :+1:

PR #223, opened 2015-05-25
Avoid timeouts when ogr2ogr queries pg_* catalog

### LINKED ISSUE #204: large exports don't work
some exports that worked before stop working, for example:

```
https://viz2.cartodb.com/api/v1/sql?q=select%20lat,long,temp,date%20from%20ow&format=csv
```
the error is:

```
"Your query was not able to finish. Either you have too many queries running or the one you are trying to run is too expensive. Try again."
```

Sounds reasonable for me but we should provide a way to download large datasets. We could use the ``COPY`` command to stream csv or pgdump files

rochoa:
Reintroduces tables= param in ogr2ogr exports

Fixes #204 

This will avoid running a heavy ogr2ogr query when the pg catalog is big

Ref https://github.com/CartoDB/CartoDB-SQL-API/commit/84c422c505391ef0e743aed2204214d4286d7e30

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,11 @@
 1.22.2 - 2015-mm-dd
 -------------------
 
+Bug fixes:
+
+ * Reintroduces tables= param in ogr2ogr exports (#204)
+   This will avoid running a heavy ogr2ogr query when the pg catalog is big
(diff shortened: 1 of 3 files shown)


rafatower:
:+1: 

Do you want me to do anything apart from the code review?

rochoa:
No really :smile:  Thanks!

PR #225, opened 2015-05-26
Uses :remote-addr instead of :req[X-Real-IP]

### LINKED ISSUE #197: Origin IP address is not shown on logs when X-Forwarded-For is not passed
For example:
```[2014-12-23 11:41:55.549] [INFO] [default] - [Tue, 23 Dec 2014 11:41:55 GMT] undefined GET undefined/api/v1/health 200 1 ms -> application/json; charset=utf-8 (undefined)```

It appears as the first "undefined" (the other one is the missing Host header). It seems to display always X-Forwarded-For, so the real IP address is never shown

fdansv:
Fixes #197 (I think)

Gets the requester's IP address via ```:remote-addr``` on the logging template. @rochoa is this an equivalent solution? I'm not that sure.

### DIFF
--- a/app/controllers/app.js
+++ b/app/controllers/app.js
@@ -77,7 +77,7 @@ var tableCache = LRU({
 var loggerOpts = {
     buffer: true,
     format: global.settings.log_format ||
-            ':req[X-Real-IP] :method :req[Host]:url :status :response-time ms -> :res[Content-Type]'
+            ':remote-addr :method :req[Host]:url :status :response-time ms -> :res[Content-Type]'
 };
 
 if ( global.log4js ) {
--- a/config/environments/development.js.example
+++ b/config/environments/development.js.example
@@ -5,7 +5,7 @@ module.exports.base_url     = '(?:/api/:version|/user/:user/api/:version)';
 // X-SQLAPI-Profile header containing elapsed timing for various
 // steps taken for producing the response.
 module.exports.useProfiler = true;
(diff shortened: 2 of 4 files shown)


rochoa:
I think this does not solve #197 because I guess that remote-addr will pick up the remote ip coming from the connecting client. However if it's there is a reverse proxy acting like client forwarding the requests, call it nginx or varnish, it will setup the `X-Forwarded-For` and it won't get into the log.

Look at this issue: https://github.com/nomiddlename/log4js-node/issues/139

fdansv:
Asked at the log4js repo, waiting for their reply to see if we just update the version or fork it.

fdansv:
@rochoa since I pushed to master the fix in our log4js fork... this should work correctly in this branch :D

rochoa:
Not exactly, this uses `cdb` branch from the fork. Check https://github.com/CartoDB/CartoDB-SQL-API/blob/1.22.2/package.json#L28

We need to merge your changes into `cdb` to get them over here :smile:

fdansv:
Aahhh cool, opening PR there then.

fdansv:
Merged, all ok?

rochoa:
That should be enough as this and windshaft-cartodb will pick up the dependency from `cdb` branch on next deploy.

fdansv:
Yeah but windshaft probably need to be modified as well so that it uses :remote-addr, right?

rochoa:
You can modify the configuration examples and default value but at the end is a configuration thing :wink:

PR #234, opened 2015-06-22
Prefixes OGR reserved fields with an underscore to prevent blank csv files

### LINKED ISSUE #184: export to csv with default column names fails
http://javi.cartodb.com/api/v1/sql?q=select%20st_astext(st_makepoint(0,0))&format=csv

fdansv:
@rochoa 

I'm not super confident about this code. Tests pass, and I've added an additional one, but I'm not familiar enough with the SQL syntax/spec to be certain that this covers all cases.

Fixes #184 

cc. @pramsey

### DIFF
--- a/app/models/formats/ogr.js
+++ b/app/models/formats/ogr.js
@@ -208,12 +208,15 @@ console.log('ogr2ogr ' + _.map(ogrargs, function(x) { return "'" + x + "'"; }).j
     if (out_format === 'CSV'){
       var pgFunctionNames = ["st_asbinary", "binarybase64", "st_asewkt", "st_asewkb", "ewkbbase64",
         "st_astext", "asbinary", "asewkt", "asewkb", "astext"];
-      var outerFunction = sql.match(/\s.*?\(/g)[0].trim().replace("(","").toLowerCase();
-      if (pgFunctionNames.indexOf(outerFunction)) {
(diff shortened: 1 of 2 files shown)


app/models/formats/ogr.js
```diff
@@ -203,6 +204,24 @@ console.log('ogr2ogr ' + _.map(ogrargs, function(x) { return "'" + x + "'"; }).j
       callback(err, out_filename);
     }
   );
+  function checkPGReserved(sql) {
```
pramsey: Might want to note in comment the purpose for this transformation? OGR will treat any column with the special names as if it is a geometry column. https://github.com/OSGeo/gdal/blob/trunk/gdal/ogr/ogrsf_frmts/pg/ogrpglayer.cpp#L1387

fdansv: True, adding it.

rochoa:
I'm not a big fan of this solution, manipulating the query with regular expressions to alias some functions feels a bit risky for me. Furthermore having in mind that adding tests for this is very difficult, when you have to check something different that the happy cases.

[ogr format is already extracting the *columns* and doing some stuff with it](https://github.com/CartoDB/CartoDB-SQL-API/blob/1.23.0/app/models/formats/ogr.js#L79-L113), do you think you can sneak the aliases somehow there? :smile: 

What do you think?

PR #243, opened 2015-08-04
Remove client from pool after error happens

### LINKED ISSUE #241: Aborted transactions due errors should be cleaned up
When a transaction fails to complete and no rollback is sent to the very same connection keeps the connection in aborted state until is closed. That prevents any other client to do more requests until connection is closed in the pool.

Sample error transaction:

```sql
BEGIN;
PREPARE _pstm AS select error;
EXECUTE _pstm;
COMMIT;
```

By closing the connection on error we avoid keeping connection with inconsistent states in the pool.

rochoa:
This help to avoid issues with transactions.

Closes #241

cc @javisantana

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,11 @@
 1.24.0 - 2015-mm-dd
 -------------------
 
+New features:
+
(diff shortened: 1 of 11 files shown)


zenitraM:
:+1:

PR #247, opened 2015-09-02
Set Last-Modified header based on affected tables

### LINKED ISSUE #101: Correctly set Last-Modified
Rather than using an arbitrary value we use a real value based on query tables.
Doing so would open up the doors for implementing cheap responses to HEAD requests thus speeding up revalidation (currently performed after one hour of persistence in cache, due to max-age).

rochoa:
Closes #101

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,9 @@
-1.24.1 - 2015-mm-dd
+1.25.0 - 2015-mm-dd
 -------------------
+
+New features:
+
+ * Set `Last-Modified` header based on affected tables (#101)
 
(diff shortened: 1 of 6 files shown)


rochoa:
cc @Kartones What do you think? Check the test suite to see if that's good enough for you.

Kartones:
:+1:  And perfectly fine to do things like querying cdb_tablemetadata instead of the user DB, from my side any "best effort" scenario is fine because being for syncs, if data caching is not perfect next run would get it anyway, so much better than no caching at all

PR #249, opened 2015-09-04
Adds a custom log4js format function to log sql query from POST requests

### LINKED ISSUE #206: Need to log POST content
We're getting very handy logs for debugging GET requests but not for POST requests. Some problems cannot be easily diagnosed without logging at least part of the request, say the first N characters.

rochoa:
Tokens in log4js are not provided with request/response for context, using a custom format function allow to access them.

In the case of express logger token functions receive request/response for context so it uses a custom token.

`SQL_QUERY_BODY_LOG_MAX_LENGTH = 2000` might seem a bit arbitrary but it tries to match which is considered a sane max length for a GET query string. That means that we won't log more than we might log during a GET request with a lengthy query string.

Closes #206

### DIFF
--- a/app/controllers/app.js
+++ b/app/controllers/app.js
@@ -74,16 +74,45 @@ var tableCache = LRU({
   maxAge: global.settings.tableCacheMaxAge || 1000*60*10
 });
 
-var loggerOpts = {
-    buffer: true,
-    format: global.settings.log_format ||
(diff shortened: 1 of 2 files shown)


rochoa:
@rafatower Do you think this is enough to close #206?

@luisbosque please, let me know if this is a problem for ES.

Thanks :smile:

rafatower:
Looks good to me

luisbosque:
:+1:

PR #250, opened 2015-09-07
Stop adding X-Cache-Channel header when no tables involved

rochoa:


rochoa:
@zenitraM what do you think?

zenitraM:
I think it makes sense :+1:

rochoa:
Thanks! :ship:ing it!

PR #261, opened 2016-01-14
Batch API 

dgaubert:
cc/ @rochoa @javisantana


app/controllers/generic_controller.js
```diff
@@ -0,0 +1,17 @@
+'use strict';
+
+var setCrossDomain = require('../utils/cross_domain');
```
rochoa: At some point I would like to move this as a middleware as in https://github.com/CartoDB/Windshaft-cartodb/blob/e4e7d6c840e16f58dc990786209425487e6b8d38/lib/cartodb/middleware/cors.js

app/controllers/job_controller.js
```diff
@@ -0,0 +1,526 @@
...
+    if (!_.isString(job_id)) {
+        return handleException(new Error("You must indicate a job id"), res);
+    }
+
+    if ( req.profiler ) {
+        req.profiler.start('sqlapi.job');
+    }
+
+    req.aborted = false;
+    req.on("close", function() {
```
rochoa: This made sense in queries endpoint as we are cancelling the query if request is cancelled before ending.

app/controllers/job_controller.js
```diff
@@ -0,0 +1,526 @@
...
+        function listJob(err, userDatabase) {
+            assert.ifError(err);
+
+            if (!userDatabase.authenticated) {
+                throw new Error('permission denied');
+            }
+
+            var next = this;
+
+            checkAborted('listJob');
```
rochoa: As I said, I would worry about aborted requests within this controller.

app/controllers/job_controller.js
```diff
@@ -0,0 +1,526 @@
...
+            if (result.host) {
+              res.header('X-Served-By-DB-Host', result.host);
+            }
+
+            res.send(result.job);
+        }
+    );
+};
+
+// jshint maxcomplexity:21
```
rochoa: :imp:

app/postgresql/error_handler.js
```diff
@@ -38,6 +38,8 @@ ErrorHandler.prototype.getStatus = function() {
 
     if (message && message.match(/permission denied/)) {
         statusError = 401;
+    } else if (message && message.match(/not found/)) {
```
rochoa: What is the use case for this?

app/services/user_database_service.js
```diff
@@ -0,0 +1,96 @@
+'use strict';
+
+var step = require('step');
+var _ = require('underscore');
+var AuthApi = require('../auth/auth_api');
+
+function UserDatabaseService() {
+}
+
+UserDatabaseService.prototype.getUserDatabase = function (options, callback) {
```
rochoa: I don't like `options` here, shouldn't we pass what we need as different params?

batch/batch.js
```diff
@@ -0,0 +1,148 @@
...
+        }
+        queue = self.jobQueuePool.createQueue(host);
+
+        // do forever, it does not throw a stack overflow
+        forever(function (next) {
+            self._consumeJobs(host, queue, next);
+        }, function (err) {
+            self.jobQueuePool.remove(host);
+
+            if (err.name === 'EmptyQueue') {
```
rochoa: Why does this need a different handling?

batch/job_backend.js
```diff
@@ -0,0 +1,316 @@
+'use strict';
+
+var uuid = require('node-uuid');
+var queue = require('queue-async');
+var JOBS_TTL_IN_SECONDS = 48 * 3600;
```
rochoa: Can we make this value configurable?

app/controllers/generic_controller.js
dgaubert: Me too, but I stopped the refactor at this point because I didn't want add more changes to this PR. There are another things that they could fit into a middleware pattern as well.

rochoa: Agree, let's move on with this.

On Thu, Jan 21, 2016 at 12:20 PM, Daniel <notifications@github.com> wrote:

> In app/controllers/generic_controller.js
> <https://github.com/CartoDB/CartoDB-SQL-API/pull/261#discussion_r50388714>
> :
>
> > @@ -0,0 +1,17 @@
> > +'use strict';
> > +
> > +var setCrossDomain = require('../utils/cross_domain');
>
> Me too, but I stopped the refactor at this point because I didn't want add
> more changes to this PR. There are another things that they could fit into
> a middleware pattern as well.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/CartoDB-SQL-API/pull/261/files#r50388714>.
>



-- 
Raul Ochoa
rochoa@cartodb.com

app/controllers/job_controller.js
dgaubert: Totally, here it doesn't make sense, I'm going to remove it

app/controllers/job_controller.js
dgaubert: +1

app/postgresql/error_handler.js
dgaubert: I think It was a attempt to handle Not Found error in Job Controller, but it doesn't make sense here because it is handled in utils/error_handler. I'm going to remove it.

app/services/user_database_service.js
dgaubert: Ok, since `checkAborted` function is not longer needed, passing four params is more readable.

batch/batch.js
dgaubert: because of logs. I don't want to be noisy. If any queue is empty I just want to log a one-line message, avoiding stack trace.

batch/job_backend.js
dgaubert: Yes, we can.

app/services/user_database_service.js
dgaubert: But query controller is also using this service. I have to keep `checkAborted` as parameter and then check if it is passed. ;)

rochoa: I would completely remove `checkAborted` from UserDatabaseService.

As I would remove anything that is not its responsibility.

From my point of view, metadataBackend should be passed when creating
the UserDatabaseService instance. And getUserDatabase should receive a user
name and an authApi instance.

On Thu, Jan 21, 2016 at 3:16 PM, Daniel <notifications@github.com> wrote:

> In app/services/user_database_service.js
> <https://github.com/CartoDB/CartoDB-SQL-API/pull/261#discussion_r50404929>
> :
>
> > @@ -0,0 +1,96 @@
> > +'use strict';
> > +
> > +var step = require('step');
> > +var _ = require('underscore');
> > +var AuthApi = require('../auth/auth_api');
> > +
> > +function UserDatabaseService() {
> > +}
> > +
> > +UserDatabaseService.prototype.getUserDatabase = function (options, callback) {
>
> But query controller is also using this service. I have to keep
> checkAborted as parameter and then check if it is passed. ;)
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/CartoDB/CartoDB-SQL-API/pull/261/files#r50404929>.
>



-- 
Raul Ochoa
rochoa@cartodb.com

dgaubert: +1

dgaubert: next step is to move it to a middleware ;)

batch/job_runner.js
```diff
@@ -0,0 +1,79 @@
+'use strict';
+
+var PSQL = require('cartodb-psql');
+var QUERY_CANCELED = '57014';
```
rochoa: Could we use this from app/postgresql/error_codes.js?

batch/job_canceller.js
```diff
@@ -0,0 +1,90 @@
...
+        }
+
+        self.jobBackend.setPending(job, callback);
+    });
+
+};
+
+JobCanceller.prototype._query = function (job, userDatabaseMetadata, callback) {
+    var pg = new PSQL(userDatabaseMetadata, {}, { destroyOnError: true });
+    var getPIDQuery = 'SELECT pid FROM pg_stat_activity WHERE query = \'' + job.query +
```
rochoa: Does this play well with queries containing quotes?

batch/job_backend.js
```diff
@@ -0,0 +1,316 @@
...
+        'updated_at',
+        'failed_reason'
+    ];
+
+    this.metadataBackend.redisCmd(this.db, 'HMGET', redisParams , function (err, jobValues) {
+        if (err) {
+            return callback(err);
+        }
+
+        function isJobFound(jobValues) {
```
rochoa: Would be better outside of this scope as it does not require *closure* scope.

test/unit/batch/username_queue.js
```diff
@@ -0,0 +1,33 @@
+var UsernameQueue = require('../../../batch/username_queue');
```
rochoa: :see_no_evil: This no longer exists, does it?

dgaubert: It was renamed to `job_queue`

batch/job_backend.js
dgaubert: Maybe as private member method... `_isJobFound`

batch/batch.js
```diff
@@ -0,0 +1,150 @@
...
+    process.on('SIGTERM', function () {
+        self.stop();
+        self.drain(function (err) {
+            if (err) {
+                console.log('Exit with error');
+                return process.exit(1);
+            }
+
+            console.log('Exit gracefully');
+            process.exit(0);
```
rochoa: I just realised about this. Not a big fan of delegating the control of SIGTERM to batch entity.

I would prefer to have app.js knowing about SIGTERM and calling batch.drain() and waiting for the callback to process.exit. Batch shouldn't know about process, IMO.

What do you think?

dgaubert: Agree! I'm going to move to app.js right away!

batch/batch.js
```diff
@@ -0,0 +1,150 @@
...
+
+    if (!job_id) {
+        return process.nextTick(function () {
+            return callback();
+        });
+    }
+
+    var queue = self.jobQueuePool.getQueue(host);
+
+    this.jobCanceller.drain(job_id, function (err) {
```
rochoa: One thing that worries me about this approach is having uninterruptible queries.

I mean, I think the logic here is good: we try to cancel the query:
 1. If we succeed to cancel, we re-enqueue it, mark it as pending, and wait until is consumed again.
 1. If we fail to cancel, we keep it running and we don't change the status.

However the query might end at some point but the job will be kept as ... `running` (?) status.

dgaubert: We'll need one status more. `unknown` to indicate to user that it wasn't possible to know what happened with the job due to cancel his job failed while draining. In this scenario, user will have to review if his job was done successfully or not. 

Do you have a better approach?

rochoa: I don't have a better approach. I agree with you on having `unknown` for these scenarios as the current alternative :+1:

rochoa:
:shipit: !

PR #266, opened 2016-01-28
Fix X-Cache-Channel generation when request are not authenticated

### LINKED ISSUE #265: X-Cache-Channel generation fails when request are not authenticated
As it is trying to retrieve last updated time for involved tables from CDB_TableMetadata, public user does not have permission to read from that table.

rochoa:
This changes UserDatabaseService to retrieve all connection params in two different objects so it is possible to run queries with the authenticated user to retrieve information from CDB_TableMetadata, but it keeps the normal behaviour for the user query.

Fixes #265

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,6 +1,10 @@
 1.25.1 - 2016-mm-dd
 -------------------
 
+Bug fixes:
+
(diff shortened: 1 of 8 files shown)


rochoa:
cc @dgaubert

dgaubert:
:+1:

PR #273, opened 2016-02-22
Use node-cartodb-querytables library and add Surrogate-Keys

zenitraM:
This moves all handling of caches to the [https://github.com/CartoDB/node-cartodb-query-tables](node-cartodb-querytables) lib. 

As a bonus, it adds [Surrogate-Key](https://github.com/CartoDB/CartoDB-SQL-API/compare/new_querytables?expand=1#diff-f9f492ee6f5bd5964aaaacbfd7fda591R180) headers to responses.


rochoa:
I put back a cache mechanism for query tables by wrapping cartodb-query-tables module and using the same LRU we were using before.

I think we are good to :shipit:.

zenitraM:
:shipit: :bomb:

PR #280, opened 2016-03-21
Batch multiquery jobs

### LINKED ISSUE #279: Batch API multiple queries
I would like to send more than one query within the same request and I expect to run the in serial. I imagine something like:

```
POST /api/v2/sql/job 
{
    "queries": [
        "CREATE TABLE table_name_2 AS SELECT * FROM table_name_1",
        "select CDB_Cartodbfytable('table_name_2')"
    ]
}
```

dgaubert:
Fixes #279

cc/ @rochoa

### DIFF
--- a/app/controllers/job_controller.js
+++ b/app/controllers/job_controller.js
@@ -221,6 +221,23 @@ JobController.prototype.getJob = function (req, res) {
     );
 };
 
+function isValidJob(sql) {
+    if (_.isArray(sql)) {
+        for (var i = 0; i < sql.length; i++) {
+            if (!_.isString(sql[i])) {
+                return false;
+            }
+        }
+        return true;
+    }
+
+    if (!_.isString(sql)) {
+        return false;
+    }
+
+    return true;
+}
+
 JobController.prototype.createJob = function (req, res) {
     var self = this;
     var body = (req.body) ? req.body : {};
@@ -228,8 +245,8 @@ JobController.prototype.createJob = function (req, res) {
     var sql = (params.query === "" || _.isUndefined(params.query)) ? null : params.query;
     var cdbUsername = cdbReq.userByReq(req);
 
-    if (!_.isString(sql)) {
-        return handleException(new Error("You must indicate a sql query"), res);
+    if (!isValidJob(sql)) {
+        return handleException(new Error("You must indicate a valid SQL query"), res);
     }
 
     if ( req.profiler ) {
@@ -300,7 +317,7 @@ JobController.prototype.updateJob = function (req, res) {
     var sql = (params.query === "" || _.isUndefined(params.query)) ? null : params.query;
     var cdbUsername = cdbReq.userByReq(req);
 
-    if (!_.isString(sql)) {
+    if (!isValidJob(sql)) {
         return handleException(new Error("You must indicate a sql query"), res);
     }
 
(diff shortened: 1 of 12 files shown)


app/controllers/job_controller.js
```diff
@@ -228,8 +245,8 @@ JobController.prototype.createJob = function (req, res) {
     var sql = (params.query === "" || _.isUndefined(params.query)) ? null : params.query;
```
rochoa: Why don't we add a `queries` option? It's a bit weird to have `query` and being able to send an array, isn't it?

dgaubert: I think is a little confusing for users to use `query` in some cases and `queries` in others. Probably, we should rename that property for something more generic like `do`, `run`, `execute` or `process` (any clues?) and users are able to send a single query or an array of queries without taking care of which property they have to use.

batch/job_backend.js
```diff
@@ -185,27 +219,33 @@ JobBackend.prototype.get = function (job_id, callback) {
     });
 };
 
-JobBackend.prototype.setRunning = function (job, callback) {
+JobBackend.prototype.setRunning = function (job, index, callback) {
+    var self = this;
     var now = new Date().toISOString();
```
rochoa: Should we use `toUTCString`?

batch/job_backend.js
```diff
@@ -236,6 +280,13 @@ JobBackend.prototype.setDone = function (job, callback) {
         'updated_at', now
     ];
 
+    if (!callback) {
+        callback = index;
+    } else if (index >= 0 && index < job.query.length) {
+        job.query[index].status = 'done';
```
rochoa: I see a lot of uses of hardcoded strings for status: I would create some constants.

batch/job_canceller.js
```diff
@@ -55,15 +73,14 @@ JobCanceller.prototype.drain = function (job_id, callback) {
...
+        self.jobBackend.setPending(job, queryIndex, callback);
     });
 
 };
 
 JobCanceller.prototype._query = function (job, userDatabaseMetadata, callback) {
     var pg = new PSQL(userDatabaseMetadata, {}, { destroyOnError: true });
-    var escapedJobQuery = pg.escapeLiteral(job.query + ' /* ' + job.job_id + ' */');
-    var getPIDQuery = 'SELECT pid FROM pg_stat_activity WHERE query = ' + escapedJobQuery;
+    var getPIDQuery = "SELECT pid FROM pg_stat_activity WHERE query LIKE '/* " + job.job_id + " */%'";
```
rochoa: Why do we need to change from `=` to `LIKE`.

rochoa:
:+1: Apart from my proposed changes, which are not blockers.

dgaubert:
Shall we create issues with proposed changes?

batch/job_backend.js
dgaubert: Shame on me :cry:

batch/job_backend.js
dgaubert: Both are in UTC time zone. `toUTCString` is more human readable but `toISOString` is better for sorting and things like that. Do we have guidelines for that?

batch/job_canceller.js
dgaubert: Well, I don't know if using `LIKE` is better in performance than `=`. But now, is checking if the query starts with `batch id` instead of checking the whole query. Also, in this way, I don't have in mind `escaped literals` thing.

PR #283, opened 2016-03-31
Improved initialization in Batch API

### LINKED ISSUE #282: Improve Batch API startup
So far, batch API has implemented a lazy startup. This method is really good when Batch API have a considerable amount of traffic, but nowadays Batch API is idle and after startup we have to implement a discover mechanism to consume job from queues.

dgaubert:
Now Batch API looks for queues with jobs to consume at service initiaization.

Fix #282 

cc @rochoa

### DIFF
--- a/batch/job_subscriber.js
+++ b/batch/job_subscriber.js
@@ -1,17 +1,15 @@
 'use strict';
 
-function JobSubscriber(redis, metadataBackend) {
+function JobSubscriber(redis, queueSeeker) {
     this.channel = 'batch:hosts';
     this.client = redis.createClient(global.settings.redis_port, global.settings.redis_host);
-    this.metadataBackend = metadataBackend;
-    this.db = 5;
-    this.redisPrefix = 'batch:queues:';
+    this.queueSeeker = queueSeeker;
 }
 
 JobSubscriber.prototype.subscribe = function (onMessage) {
     var self = this;
 
-    self._lookForQueues(onMessage, function (err) {
+    self.queueSeeker.seek(onMessage, function (err) {
         if (err) {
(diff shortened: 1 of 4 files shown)


batch/job_subscriber.js
```diff
@@ -1,15 +1,60 @@
...
+        }
+
+        var queues = currentCursor[1];
+
+        if (!queues) {
+            return callback(null);
+        }
+
+        queues.forEach(function (queue) {
+            var host = queue.substr(queue.lastIndexOf(':') + 1);
```
rochoa: Isn't this the same as `this.redisPrefix.length`?

batch/job_subscriber.js
```diff
@@ -1,15 +1,60 @@
...
+
+        self.client.subscribe(self.channel);
+        self.client.on('message', onMessage);
+    });
 };
 
+JobSubscriber.prototype._lookForQueues = function (onMessage, callback) {
+    var pattern = this.redisPrefix + '*';
+
+    scanQueues(this.metadataBackend, this.db, ['0'], pattern, this.channel, onMessage, callback);
```
rochoa: Is the `['0']` format required because how the redis client works?

batch/job_subscriber.js
dgaubert: Oh! Good catch! ;)

batch/job_subscriber.js
dgaubert: Yes, it is. A little weird...

dgaubert:
I've just refactored in order to improve isolation and maintainability. Please, check it out!

PR #287, opened 2016-04-18
updated note and added link to client examples

csobier:
For [Docs issue#784](https://github.com/CartoDB/docs/issues/784).

Thanks for the resources @dgaubert !  See the updates I applied, I updated the note and added a section at the end about complete examples. This is only one option, let me know if you rather I not use your GIST link, and include these locally within each operation?


dgaubert:
Cool!

I rather not use my GIST, I think is better to include the examples, thus user can read the whole documentation in just one place.

Thanks!

csobier:
Totally agree- I'm on it!

dgaubert:
👍

PR #288, opened 2016-04-18
Batch apply query size limit

dgaubert:
cc @rochoa


app/controllers/job_controller.js
```diff
@@ -318,7 +342,11 @@ JobController.prototype.updateJob = function (req, res) {
     var cdbUsername = cdbReq.userByReq(req);
 
     if (!isValidJob(sql)) {
-        return handleException(new Error("You must indicate a sql query"), res);
+        return handleException(new Error('You must indicate a valid SQL'), res);
+    }
+
+    if (reachMaxQuerySizeLimit(sql)) {
+        return handleException(new Error('Query is too long'), res);
```
rochoa: I would give some feedback on what is the limit and how much the query is exceeding it.

rochoa:
Let's 🚢  🇮🇹  once we get a better feedback in the error message 😈 .

dgaubert:
I've just done some changes... 👀 please!

PR #301, opened 2016-05-17
Job model refactor

dgaubert:
**This PR is the previous step to get done fallback jobs**. So far, jobs were literal objects that were passed among runner, backend and so on. After implement multiple query jobs and fallbacks the code was ugly and turned hard to maintain. I've decided redesign jobs, isolating them into classes and every job class implements a common interface to be handled by runner without taking care about what kind of job is. Thus, we can add any new types of jobs easily. Also we improve the maintainability and readability. 

There is some improvements done in fallback implementation, this PR aims to show what I had to change to get done fallbacks.

cc @rochoa


app/app.js
```diff
@@ -183,15 +185,17 @@ function App() {
...
+
 
     var genericController = new GenericController();
     genericController.route(app);
 
     var queryController = new QueryController(userDatabaseService, tableCache, statsd_client);
     queryController.route(app);
 
-    var jobController = new JobController(userDatabaseService, jobBackend, jobCanceller);
+    var jobController = new JobController(userDatabaseService, jobService, jobCanceller);
```
rochoa: I understand job canceller is no longer needed because it's exposed through job service, right?

app/controllers/job_controller.js
```diff
@@ -37,12 +37,13 @@ function getMaxSizeErrorMessage(sql) {
     );
 }
 
-function JobController(userDatabaseService, jobBackend, jobCanceller) {
+function JobController(userDatabaseService, jobService) {
```
rochoa: Way better to have just one clean/clear dependency here 👍 .

app/app.js
dgaubert: right!

app/controllers/job_controller.js
dgaubert: this refactor helped to me to clean and get cleared several things. ☺️

batch/batch.js
```diff
@@ -46,11 +48,51 @@ Batch.prototype._subscribe = function () {
...
+            return callback(emptyQueueError);
+        }
+
+        self.jobQueuePool.setCurrentJobId(host, job_id);
+
+        self.jobRunner.run(job_id, function (err, job) {
+            self.jobQueuePool.removeCurrentJobId(host);
+
+            if (err && err.name === 'JobNotRunnable') {
+                console.log(err.message);
```
rochoa: Avoid `console` usage, favor `debug` if you need it.

batch/job_backend.js
```diff
@@ -27,155 +19,52 @@ function JobBackend(metadataBackend, jobQueueProducer, jobPublisher, userIndexer
...
-        if (job.status !== jobStatus.PENDING) {
-            return callback(new Error('Job is not pending, it cannot be updated'));
-        }
-
-        if (Array.isArray(job.query)) {
-            for (var i = 0; i < job.query.length; i++) {
-                if (job.query[i].status !== jobStatus.PENDING) {
-                    return callback(new Error('Job is not pending, it cannot be updated'));
-                }
+JobBackend.prototype.toRedisParams = function (data) {
```
rochoa: I would avoid making this _pure_ functions part of the prototype.

batch/job_backend.js
```diff
@@ -27,155 +19,52 @@ function JobBackend(metadataBackend, jobQueueProducer, jobPublisher, userIndexer
...
-        var initialLength = job_ids.length;
+JobBackend.prototype.toObject = function (job_id, redisParams, redisValues) {
+    var obj = {};
 
-        self._getCleanedList(username, job_ids, function (err, jobs) {
-            if (err) {
-                return callback(err);
-            }
+    redisParams.shift(); // job_id value
+    redisParams.pop(); // WARN: weird function pushed by metadataBackend
```
rochoa: What is the weird function?

dgaubert: I think is a callback bound in someway. 

```
function (err, data) {
                    if (logSlowQueries) {
                        var elapsedTime = Date.now() - executeStartTime;
                        if (elapsedTime > slowQueriesElapsedThreshold) {
                            console.log(
                                JSON.stringify({db: db, action: 'query', commands: commands, elapsed: elapsedTime})
                            );
                        }
                    }
                    if (redisClient) {
                        redis_pool.release(db, redisClient);
                    }
                    callback(err, data);
                }
```

batch/job_backend.js
```diff
@@ -186,244 +75,175 @@ JobBackend.prototype.get = function (job_id, callback) {
         'query',
         'created_at',
         'updated_at',
+        'host',
         'failed_reason'
     ];
 
-    this.metadataBackend.redisCmd(this.db, 'HMGET', redisParams , function (err, jobValues) {
+    this.metadataBackend.redisCmd(this.db, 'HMGET', redisParams , function (err, redisValues) {
```
rochoa: I recommend you to use `getMultipleUserDBParams` here, where you can pass a map of the keys you want and the resulting `keys=>values` you want.

So you pass:

```js
{
  wanted_key_1: 'desired_target_key_1',
  wanted_key_2: 'desired_target_key_1' 
}
```

And you get:

```js
{
  desired_target_key_1: 'value_1',
  desired_target_key_2: 'value_2' 
}
```

batch/job_backend.js
dgaubert: yep, it isn't needed

batch/job_backend.js
```diff
@@ -186,244 +75,175 @@ JobBackend.prototype.get = function (job_id, callback) {
...
             return callback(err);
         }
 
-        self.get(job.job_id, callback);
+        self.save(data, callback);
     });
 };
 
-JobBackend.prototype.setDone = function (job, index, callback) {
+JobBackend.prototype.save = function (data, callback) {
```
rochoa: Can we use another param name? `job`?

Applies to previous methods.

batch/job_factory.js
```diff
@@ -0,0 +1,24 @@
+'use strict';
+
+var JobSimple = require('./job_simple');
```
rochoa: Probably we need to start with some kind of namespacing here:

```js
var models = require('./model')
```

batch/job_simple.js
```diff
@@ -0,0 +1,29 @@
+'use strict';
```
rochoa: Did you consider transforming a "job_simple" into a "job_multiple"?

batch/job_base.js
```diff
@@ -0,0 +1,137 @@
...
+    this.data.query = query;
+};
+
+JobBase.prototype.setStatus = function (finalStatus) {
+    var now = new Date().toISOString();
+    var initialStatus = this.data.status;
+    var isValid = this.isValidStatusTransition(initialStatus, finalStatus);
+
+    if (!isValid) {
+        throw new Error('Cannot set status from ' + initialStatus + ' to ' + finalStatus);
```
rochoa: What about current errors? Is this a backwards incompatible change?

We already have these messages in the public documentation: http://docs.cartodb.com/cartodb-platform/sql-api/sql-batch-api/#cancel-a-job

dgaubert: Error messages implies a backwards incompatible change?. I did not know it.

batch/job_simple.js
dgaubert: I fact, I did. But "job_multiple" changes the job structure from this:
```
"query": [
    "SELECT * FROM untitle_table_4 limit 1",
    "SELECT * FROM untitle_table_4 limit 2"
]
```
to: 
```
"query": [{
    query:  "SELECT * FROM untitle_table_4 limit 1",
    status: "pending"
}, {
    query:  "SELECT * FROM untitle_table_4 limit 2",
    status: "pending"
}]
```
and it would be a backward incompatible for "job_simple". I think.

batch/job_backend.js
dgaubert: 😞  

I can't use it, `getMultipleUserDBParams` uses internally `rails:users:*` as key, but batch works under `batch:*` key.

Anyway, thanks for sharing, it will be useful for sure!

PR #302, opened 2016-05-18
Batch fallback job

dgaubert:
**Note**: added a new dev-dependency: `supertest` due to our custom `assert.response` is not able to parse nested object as `querystring`. We should take a look of that as soon as possible.

cc @rochoa


dgaubert:
ready again for CR 😄

rochoa:
What is exactly not supported in `assert.response` method?

dgaubert:
`assert.response` uses `querystring` module to parse `request.body` for POST method (among others). So, fallback jobs (with nested objects) is not parsed properly and the tests were not working properly. I had two options, first one, use `supertest` and the second one, fix `assert.response` using, for example [qs](https://github.com/ljharb/qs). What do you prefer?

rochoa:
What I don't want is to have two different interfaces to do tests because we will end with a mix of both. What's is an example of the problem you're commenting?

We had a similar assert.response in windshaft-cartodb but was modified to use request: https://github.com/CartoDB/Windshaft-cartodb/blob/eb091caf4a7348e3f24b94d09dac8255066200bb/test/support/assert.js#L81

The problem with that is that is targeting express 4.x, so it won't be a plug&play replacement.

dgaubert:
Agree, mixing interfaces is not a good idea, but I chose to use `supertest` for two reasons:
    - Is _de-facto_ standard library for those things. 
    - I was busy trying to resolve status transitions with fallback jobs and I didn't want to fix another one because of test.

In fact, I think that our `assert.response` is sending `request.body` as URL query-string params for POST method.

Then, I'm going to fix `assert.response` and modify fallback test.

dgaubert:
I was totally wrong, finally I've modified fallback's tests to use our `assert.response`. I had only to use `qs`module (supports nested objects) to parse the request's body.

:shame:

batch/models/index.js
```diff
@@ -2,5 +2,6 @@
 
 var JobSimple = require('./job_simple');
 var JobMultiple = require('./job_multiple');
+var JobFallback = require('./job_fallback');
 
-module.exports = [ JobSimple, JobMultiple ];
+module.exports = [ JobSimple, JobMultiple, JobFallback ];
```
rochoa: Isn't a bit hacky to export different models in an array? I understand it's convenient due the usage we do, but I don't think it's the best way to make then available.

batch/models/job_fallback.js
```diff
@@ -0,0 +1,292 @@
...
+
+        if (typeof query.query[i].query !== 'string') {
+            return false;
+        }
+    }
+
+    return true;
+};
+
+// jshint maxcomplexity: 10
```
rochoa: This should be inside the function block, otherwise it's applied to the rest of the file.

test/acceptance/job.fallback.test.js
```diff
@@ -0,0 +1,1383 @@
...
+                done();
+            });
+        });
+
+        it('job should be done', function (done) {
+            var expectedQuery = {
+                "query": [{
+                    "query": "SELECT * FROM untitle_table_4",
+                    "onsuccess": "SELECT * FROM untitle_table_4 limit 1",
+                    "status": ["done", "done"]
```
rochoa: Why do we have two `done` here?

batch/models/job_fallback.js
```diff
@@ -0,0 +1,292 @@
...
+    jobStatus.CANCELLED,
+    jobStatus.DONE,
+    jobStatus.FAILED,
+    jobStatus.UNKNOWN
+];
+function isFinalStatus(status) {
+    return finalStatus.indexOf(status) !== -1;
+}
+
+function JobFallback(data) {
```
rochoa: I still don't like the `data` naming here, what about jobDefinition or something that at least states it's a job request.

test/acceptance/job.fallback.test.js
rochoa: I guess the second one is for the `onsucess` _callback_. But what if I send a job with:

```
[{
  query: 'select 1 this will fail',
  onsucess: 'select 2',
  onerror: 'select 3'
}]
```

What is the expected output? And what if I just flip callbacks, as in:

```
[{
  query: 'select 1 this will fail',
  onerror: 'select 3',
  onsucess: 'select 2'
}]
```

In general, does it make sense to include the status of the _callbacks_ in the results? For me it would be enough if my query succeeded or failed, and then have the guarantee that the fallbacks run adequately.

dgaubert: When a query has fallback: first status is referencing to query and the second one is referencing to query's fallback.

I took this approach for two reasons, first one to give to the user more info about how is performing his job and second one, batch needs to know fallback's status to determine job transitions between statuses.

dgaubert: a job with `onsuccess` and `onerror` fallbacks only is going to run one of them, never both. The second status depends on the first one (the query)
if `success` -> `onsuccess`;
if `error` -> `onerror`;

dgaubert: Probably there is a way to avoid expose fallback statuses, but I though from point of view of the user is worth to know also what's happened with his fallbacks.

rochoa: For me this is exposing an implementation detail to the end user: the user doesn't need to know anything about how we determine the job transition.

So ... if we are just giving a second status... would be better to have a `fallback_status` property? My point is: as a user I don't want to check `status` as {Array} or {String} depending on what I sent, or have to check if status is an array or an string.

dgaubert: The main reason to expose that is to give feedback to the end user, apart from that, we use these status to know if the job has more queries to run or not.

dgaubert: With the second point I totally agree, in fact, I thought about that. But first, I needed to make everything working together.

I think, is time to for that.

dgaubert:
I've done requested changes, please CR 🙏 🙏 🙏

batch/models/job_fallback.js
```diff
@@ -73,8 +73,8 @@ JobFallback.is = function (query) {
     return true;
 };
 
-// jshint maxcomplexity: 10
 JobFallback.prototype.init = function () {
+    // jshint maxcomplexity: 10
```
rochoa: I would try to adjust the value to the actual one, so in the future it gets worse it gets triggered.

rochoa:
👍  The status is much nicer this way, isn't it? 😈 

:shipit: !!

dgaubert:
Yes, I think so! 🐑

PR #305, opened 2016-05-24
Payload size validates multiple queries and fallback queries

rochoa:
As far as I understand we will store almost all the body in most of the cases, except a small part of it: the root `{query: ''}`. So to keep it simple I decided to check against the full body instead of just what is in the query argument.

This will also work with fallback queries.

It uses a middleware to check the body size.

cc @dgaubert


dgaubert:
Oh! middlewares... I love them 😍 😍 😍 

👍

PR #308, opened 2016-05-25
Fixed issue with status transition fallback jobs

dgaubert:
After some tests with camshaft integration I realised that status transition is not working well. This PR fix those issues.

Please @rochoa, could you CR?

### DIFF
--- a/test/acceptance/job.fallback.test.js
+++ b/test/acceptance/job.fallback.test.js
@@ -875,6 +875,81 @@ describe('Batch API fallback job', function () {
         });
     });
 
+    describe('"onerror" should not be triggered for any query', function () {
+        var fallbackJob = {};
+
+        it('should create a job', function (done) {
+            assert.response(app, {
+                url: '/api/v2/sql/job?api_key=1234',
+                headers: {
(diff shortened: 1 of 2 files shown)


test/acceptance/job.fallback.test.js
```diff
@@ -875,6 +875,81 @@ describe('Batch API fallback job', function () {
...
+            });
+        });
+
+        it('job should be failed', function (done) {
+            var expectedQuery = {
+                query: [{
+                    query: 'SELECT * FROM untitle_table_4 limit 1',
+                    onerror: 'SELECT * FROM untitle_table_4 limit 2',
+                    status: 'done',
+                    fallback_status: 'pending'
```
rochoa: Should we have another status for these cases? I mean, the fallback_status is pending but it's actually not because is not gonna be run. What about using: "skipped" or "ignored"?

dgaubert: Yes, we should. "skipped" sounds great. Do you want it to be implemented in this PR?

rochoa: No, I was just wondering. So go ahead with this one and :shipit:.

PR #311, opened 2016-05-26
Fixes #309, added skipped status to fallback-jobs

### LINKED ISSUE #309: "skipped" fallback_status for query callback
If a query callback, onerror or onsuccess, is not not executed because the other condition applied the status should be "skipped".

dgaubert:
Please @rochoa could you CR?

PS: We should do the same for multiple jobs, shouldn't we?


batch/models/job_fallback.js
```diff
@@ -185,9 +185,79 @@ JobFallback.prototype.setStatus = function (status, errorMesssage) {
...
+
+    if (status === jobStatus.CANCELLED || status === jobStatus.FAILED) {
+        this._setRestPendingToSkipped(status);
+    }
+};
+
+JobFallback.prototype._setSkippedQueryStatus = function () {
+    // jshint maxcomplexity: 8
+    for (var i = 0; i < this.data.query.query.length; i++) {
+        if (this.data.query.query[i].status === jobStatus.FAILED && this.data.query.query[i].onsuccess) {
```
rochoa: I'm fine going with this approach but, and I'm throwing this without a lot of understanding of all implications, so sorry I'm too wrong, I'm going to add:

Aren't we implementing a lot of the things a FSM solves? I'm worried about the complexity of these classes, almost all their methods are getting more complex over time, and I was wondering if we could use a FSM to help with it.

```
if (this.data.query.query[i].status === jobStatus.DONE && this.data.query.query[i].onerror) {
    if (this.isValidStatusTransition(this.data.query.query[i].fallback_status, jobStatus.SKIPPED)) {
        this.data.query.query[i].fallback_status = jobStatus.SKIPPED;
    }
}
```

vs

```
this.job.setStatus(jobStatus.FAILED, i);
```

I guess, most likely, that would require to transform a bit the `data` dummy structure into something that also is a FSM, so each query can implement their transitions.

dgaubert: Yep, I'm also worried about the complexity of these classes are getting. I think, I'm messing up the FSM and State Pattern. It's working even though the implementation can be improved. I'm going to give it some thought.

dgaubert:
Please, @rochoa could you CR again?

PS: I'm fine with this approach, but there are some details that can be smoothed out.

batch/models/job_state_machine.js
```diff
@@ -0,0 +1,46 @@
...
+var jobStatus = require('../job_status');
+var finalStatus = [
+    jobStatus.CANCELLED,
+    jobStatus.DONE,
+    jobStatus.FAILED,
+    jobStatus.UNKNOWN
+];
+
+var validStatusTransitions = [
+    [jobStatus.PENDING, jobStatus.RUNNING],
```
rochoa: What about having:

```
var validStatusTransitions = [
    [jobStatus.PENDING, [jobStatus.RUNNING,jobStatus.SKIPPED, jobStatus.CANCELLED, jobStatus.UNKNOWN]],
    [jobStatus.RUNNING, [jobStatus.DONE, jobStatus.FAILED, jobStatus.CANCELLED, jobStatus.PENDING, jobStatus.UNKNOWN]]
];
```

rochoa: Not that important BTW.

rochoa:
It's way better, isn't? 😄

batch/models/job_state_machine.js
dgaubert: Nice!

dgaubert:
Yes, it is. 😆

batch/models/job_state_machine.js
dgaubert: I've just tried your suggestion: 
```
    for (var i = 0; i < validStatusTransitions.length; i++) {
        for (var j = 0; j < validStatusTransitions[i][1].length; j++) {
            try {
                assert.deepEqual(transition, [validStatusTransitions[i][0], validStatusTransitions[i][1][j]]);
                return true;
            } catch (e) {
                continue;
            }
        }
    }
```
Complexity increases a little, doesn't it?

PR #312, opened 2016-05-26
Adds profiling to batch and job controller

dgaubert:
cc @rochoa


batch/job_runner.js
```diff
@@ -2,17 +2,22 @@
...
     this.queryRunner = queryRunner;
     this.userDatabaseMetadataService = userDatabaseMetadataService; // TODO: move to queryRunner
+    this.statsdClient = statsdClient;
 }
 
 JobRunner.prototype.run = function (job_id, callback) {
     var self = this;
 
+    self.profiler = new Profiler({ statsd_client: self.statsdClient });
+    self.profiler.start('sqlapi.batch.' + job_id);
```
dgaubert: if I'm not wrong, we can aggregate by whole batch stats and by job, right?

rochoa: I would not track time per job id, specially because we will be creating too many db file in graphite.

app/controllers/job_controller.js
```diff
@@ -97,21 +98,33 @@ JobController.prototype.cancelJob = function (req, res) {
...
+                req.profiler.sendStats();
 
-            if (global.settings.api_hostname) {
-              res.header('X-Served-By-Host', global.settings.api_hostname);
+                res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());
             }
 
-            if (result.host) {
-              res.header('X-Served-By-DB-Host', result.host);
+            if (self.statsdClient) {
```
rochoa: Cannot we warrantee statsClient is always present, or something with the same interface so we don't have to keep checking we have one instance?

app/controllers/job_controller.js
```diff
@@ -97,21 +98,33 @@ JobController.prototype.cancelJob = function (req, res) {
             });
         },
         function handleResponse(err, result) {
+            // jshint maxcomplexity: 8
```
rochoa: It cannot be so common to introduce a bump for max complexity, otherwise it's better to not have the rule from the beginning.

dgaubert:
@rochoa could you 👀 , 🙏 🙏 🙏

PR #321, opened 2016-06-29
Adds start and end time for batch queries with fallback

### LINKED ISSUE #320: Batch API should report start and end time for each query

rochoa:
Closes #320.


dgaubert:
nice! 👍

PR #323, opened 2016-06-29
Now Batch API broadcast to other APIs every time that re-enqueues a job

dgaubert:
After landing the whole family of multi-query jobs (fallbacks and so on) and in order improve query distribution among APIs, now Batch API broadcast to the PUB/SUB channel every time that re-enqueues a multi-query job. Thus, we avoid scenarios that only one API consumes a multiple job with 100 queries.


rochoa:
🚢 🇮🇹

PR #326, opened 2016-07-07
Fix unpublished messages in Batch Publisher

dgaubert:
- Now batch publisher sends a ping to server before publishing and create a new connection if error.
 - Batch publisher and subscriber logs (if debug enabled) both outcoming and incoming messages to give more visibility.

cc @rochoa


rochoa:
Let's :shipit: !

dgaubert:
\o/

PR #348, opened 2016-08-25
added create indexes to sql api docs

csobier:
I think this legacy editor content (creating indexes) is better suited to be located directly with the SQL API docs. (I also fixed a broken hyperlink from the Query Optimization section).

@dgaubert , do you approve?


dgaubert:
👍

PR #353, opened 2016-08-30
Remove batch list update

rochoa:
## TODO

- [x] Remove patch/put endpoints for jobs.
- [x] Remove list endpoint.
- [x] Log job creation, per user.

cc @dgaubert


dgaubert:
Nice. Removing code is the greatest pleasure of our work. 😄 
 
👍

PR #359, opened 2016-09-13
added another best practice to batch queries content

### LINKED ISSUE #358: Be more clear in the docs that Batch Queries are supposed to
Batch Queries are supposed to manipulate and create new data, things like creating expensive indexes, updates over big tables, and creating tables from complex queries. For instance, doing a "select ..." query has no much effect as the results are not returned/stored in any place.

csobier:
Fixes #358 

@dgaubert , can you review and make sure it includes everything you were thinking of? Also, not sure where the .DS_Store file came from in this PR?


doc/batch_queries.md
```diff
@@ -482,6 +482,8 @@ In some scenarios, you may need to fetch the output of a job. If that is the cas
 
 For best practices, follow these recommended usage notes when using Batch Queries:
 
+- Batch Queries are recommended for INSERT, UPDATE, and CREATE queries that manipulate and create new data, such as creating expensive indexes, applying updates over large tables, and creating tables from complex queries. Batch queries will **not** improve processing times for SELECT queries that retrieve data but do not store the results in a table
```
dgaubert: Batch queries has no effect for SELECT queries that retrieve data but do not store the results in a table. For example, the data that returns the following query `select * from my_dataset` will be lost.

Makes sense?

dgaubert:
Be careful with `.DS_Store` file. It's a MacOS file and we can't push it to the repo.

doc/batch_queries.md
csobier: Okay, how about this.
 Batch queries have no effect for SELECT queries that retrieve data but do not store the results in a table. For example, `SELECT * from my_dataset` will not produce any results.

If thats correct, I'll remove the erroneous note from other doc sections too.

csobier:
Yes, makes sense.

Weird, I have no clue how that .DS_store file got here and can't seem to delete it?

dgaubert:
I've just removed this file.

You need to configure git globally, for that, open a terminal and run:

```
echo .DS_Store >> ~/.gitignore_global
```

and finally:

```
git config --global core.excludesfile ~/.gitignore_global
```

dgaubert:
👍

PR #368, opened 2016-09-29
Implement batch logger to log query times

dgaubert:
Implement batch logger to log query times when queries are defined with id

### DIFF
--- a/batch/batch-logger.js
+++ b/batch/batch-logger.js
@@ -1,0 +1,86 @@
+'use strict';
+
+var bunyan = require('bunyan');
+var fs = require('fs');
+
+var debug = require('./util/debug')('batch-logger');
+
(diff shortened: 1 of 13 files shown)


batch/batch-logger.js
```diff
@@ -0,0 +1,85 @@
...
+    var queries = job.data.query.query;
+
+    for (var i = 0; i < queries.length; i++) {
+        var query = queries[i];
+
+        if (!query.id) {
+            continue;
+        }
+
+        var node = parseQueryId(query.id);
```
dgaubert: @rochoa, do you think It's a good idea to parse here the id? Or should it be done in beaver/logstash?

The idea is to implement logger agnostic as much as possible.

PR #384, opened 2016-11-11
Change jobs work in progress endpoint

dgaubert:
CR: @rochoa

### DIFF
--- a/app/controllers/job_controller.js
+++ b/app/controllers/job_controller.js
@@ -49,11 +49,6 @@ JobController.prototype.route = function (app) {
         this.createJob.bind(this)
     );
     app.get(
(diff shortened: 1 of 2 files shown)


app/controllers/job_controller.js
```diff
@@ -63,6 +58,10 @@ JobController.prototype.route = function (app) {
         userMiddleware, authenticatedMiddleware(this.userDatabaseService),
         this.cancelJob.bind(this)
     );
+    app.get(
+        global.settings.base_url + '/jobs-work-in-progress',
+        this.listWorkInProgressJobs.bind(this)
```
rochoa: This is missing authentication.

rochoa:
Do we want to expose this endpoint?

PR #386, opened 2016-11-29
Respect host and port settings for postgres and redis config

jchamberlain:
This PR makes sure the correct Redis and Postgres instances are prepared for testing.

I pulled down the repo for the first time the other day and tried to run tests, but many of the tests weren't connecting to the correct databases. My Postgres and Redis were on a separate host, so I had set the appropriate config in `config/environments/test.js`, but—strangely—that only helped some of the tests. Turns out `db_host` was forgotten in one place of the preparation script, and `redis_host` was not used at all. This commit fixes both.

### DIFF
--- a/test/prepare_db.sh
+++ b/test/prepare_db.sh
@@ -105,36 +105,37 @@ fi
 
 if test x"$PREPARE_REDIS" = xyes; then
 
+  REDIS_HOST=`node -e "console.log(require('${TESTENV}').redis_host || '127.0.0.1')"`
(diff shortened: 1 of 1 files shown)


rochoa:
Thanks a lot @jchamberlain!

vhamer:
I'm noting here that we have just received @jchamberlain 's signed CLA. So this PR and future are ok for commit. Thanks again for your contribution @jchamberlain !

Issue #239, opened 2015-08-03
parameterize queries 


calvinmetcalf:
the ability to parameterize queries would be extremely helpful, something like

```
?q=select foo from bar where bat=? and baz = ?&param[0]=$something$fiz$something$&param[1]=9
```

currently I've been able to do a somewhat hacky method of parameterizing via prepared statements but I got some feedback from you guys that that could lead to issues with pgbouncer

PR #387, opened 2016-11-30
Accept parameterized queries

jchamberlain:
Addresses #239 by accepting an array called `params` in the query string, e.g.,

```
{user}.carto.com/api/v2/sql/?api_key=XXXX&sql=SELECT * FROM table WHERE num > $1&params[]=5
// runs SELECT * FROM table WHERE num > 5
```

The important work was already done by @calvinmetcalf in CartoDB/node-cartodb-psql#6, so this is just a 3-line change plus tests.

Currently waiting to hear back about the CLA.


jchamberlain:
CLA signed.

jchamberlain:
Any chance this could get reviewed? (@rochoa?)

PR #390, opened 2016-12-08
fix to prevent reaching file name length limit

mbektas:


javisantana:
cc @rochoa cold you review (I think the test failing is not due to this change)

rochoa:
Code LGTM 👍 . Tests failing seems related to connectivity problems between Travis and nodejs.org, making impossible to install the correct Node.js version to run the tests. Let's wait until that problem is solved, and we can validate all tests.

rochoa:
Do you mind upgrading your branch with master? We recently fixed some unstable tests.

mbektas:
@rochoa have been using cdb branch. How is that different from master? What is the cdb for?

rochoa:
cdb branch is a stable branch, we merge changes from master into cdb once in a while, usually when we consider master good/tested enough.

I just updated `cdb` with latest improvements in testing environment/tests. So you can pick the changes from there also.

Issue #388, opened 2016-12-05
Add Golang client to libraries list?


ztephm:
hi @rochoa should i assign elsewhere?
cc @csobier 

In our SQL API docs we have a section on [libraries in different languages](https://carto.com/docs/carto-engine/sql-api/libraries-support/#libraries-in-different-languages).

S/B client is requesting we add [CARTO client for Golang](https://github.com/agonzalezro/cartodb_go). 

Do we want to add? Or need me to ask for more details?

rochoa:
Feel free to open a PR.

However, we should state clearly that all those libraries are externally developed and maintained.

ztephm:
Thanks! 

Hi @csobier I can make the PR but which path is the proper 07-libraries-support.md in? I only see one listing the libraries out in the old version of repo.

csobier:
This is the path:  https://github.com/CartoDB/CartoDB-SQL-API/blob/master/doc/libraries_support.md

Thanks! And you can tag me to review, I'll make sure the note about using external libraries is strong. It's a separate step to pull the changes into the Docs repo.

PR #391, opened 2016-12-08
added Golang client

ztephm:
Look ok @csobier ? If we need different description I can ask support client.

Closes https://github.com/CartoDB/CartoDB-SQL-API/issues/388


csobier:
Thanks @ztephm , just added the note about using external libraries, and applied our corporate branding (even if the linked repo is out-of-date).  Thank you!

For some reason, the builds are failing (not from anything we are doing). @rochoa, can you let us know when we can retest and merge?

PR #403, opened 2017-01-11
Avoid gpkg fid column

javitonino:
Fixes https://github.com/CartoDB/cartodb/issues/9559

This adds `FID = cartodb_id` to the ogr command line for gpkg. The options is documented http://www.gdal.org/drv_geopackage.html. By default, it creates a new column (`fid`) to use as a primary key of the sqlite table. With this option, we tell ogr that it can use `cartodb_id` as a primary key instead (i.e: cartodb_id is numeric and has no duplicates). That avoids creating a new column in the exported file.

CR @rochoa (old-school review assignments). I know this skips all Kanban, but we can consider this an open source contribution. Pretty much I got tired of the fid column and decided to fix it at home.


rochoa:
Nice 👍 .

Thank you for your OSS contribution 😂 .

PR #414, opened 2017-03-30
Make active GC cycles for Node.js >=v6

rochoa:


app.js
```diff
@@ -80,7 +81,17 @@ if ( ! global.settings.base_url ) {
...
-var server = require('./app/server')();
+var StatsClient = require('./app/stats/client');
+if (global.settings.statsd) {
+    // Perform keyword substitution in statsd
+    if (global.settings.statsd.prefix) {
+        var hostToken = os.hostname().split('.').reverse().join('.');
+        global.settings.statsd.prefix = global.settings.statsd.prefix.replace(/:host/, hostToken);
+    }
+}
+global.statsClient = StatsClient.getInstance(global.settings.statsd);
```
dgaubert: Do we use the global reference in another place?. It looks like we pass `statsClient` to every module that uses it.

app/stats/client.js
```diff
@@ -0,0 +1,73 @@
+var _ = require('underscore');
+var debug = require('debug')('windshaft:stats_client');
+var StatsD = require('node-statsd').StatsD;
+
+module.exports = {
```
dgaubert: We have this portion of code in other services (tiler?). We should start thinking to _modularize_ it

app/server.js
```diff
@@ -49,8 +47,8 @@ process.env.PGAPPNAME = process.env.PGAPPNAME || 'cartodb_sqlapi';
 // override Date.toJSON
 require('./utils/date_to_json');
 
-// jshint maxcomplexity:12
-function App() {
+// jshint maxcomplexity:9
```
dgaubert: ❤️

app/stats/client.js
rochoa: Yes, we have it in windshaft-cartodb. I agree, but I prefer duplication over a wrong abstraction. Let's do it the next time we do the same.

dgaubert: fair enough

PR #415, opened 2017-03-31
Avoid scan command to discover queues

### LINKED ISSUE #344: Avoid scan approach to discover active queues
After integrating successfully `redis-pool` for `job publisher`, it'd be nice to implement the same approach removing the current interval to refresh the connections to PUB/SUB.

dgaubert:
Fixes #344


batch/job_queue.js
```diff
@@ -32,7 +38,23 @@ JobQueue.prototype.size = function (user, callback) {
 };
 
 JobQueue.prototype.dequeue = function (user, callback) {
-    this.metadataBackend.redisCmd(QUEUE.DB, 'RPOP', [ QUEUE.PREFIX + user ], function(err, jobId) {
+    var dequeueScript = [
+        'local job_id = redis.call("RPOP", KEYS[1])',
```
rochoa: 😈  Nice.

dgaubert:
Yep, I realized that when I started to think about the deployment of this. I would keep the scan at startup of the service for a while. Once we are happy with it, remove the scan thing.

dgaubert:
Oh men! Thanks a lot for that test, actually is the right one!

❤️

PR #418, opened 2017-04-05
Make zip command path configurable

### LINKED ISSUE #413: zip command used in shapefile export should be configurable
It may be located in a non-standard place

It has a `FIXME` warning in the code :smile: https://github.com/CartoDB/CartoDB-SQL-API/blob/master/app/models/formats/ogr/shp.js#L36

dgaubert:
Closes #413

### DIFF
--- a/app/models/formats/ogr/shp.js
+++ b/app/models/formats/ogr/shp.js
@@ -59,15 +59,24 @@ ShpFormat.prototype.toSHP = function (options, callback) {
 
       var child = spawn(zip, ['-qrj', zipfile, outdirpath ]);
 
-      child.on('exit', function(code) {
-        //console.log("Zip complete, zip return code was " + code);
-        if (code) {
-          next(new Error("Zip command return code " + code));
-          //res.statusCode = 500;
-        }
-        next(null);
+      var stderrData = [];
+      child.stderr.setEncoding('utf8');
+      child.stderr.on('data', function (data) {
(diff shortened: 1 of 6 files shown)


app/models/formats/ogr/shp.js
```diff
@@ -63,7 +63,7 @@ ShpFormat.prototype.toSHP = function (options, callback) {
         //console.log("Zip complete, zip return code was " + code);
         if (code) {
           next(new Error("Zip command return code " + code));
-          //res.statusCode = 500; 
+          //res.statusCode = 500;
```
ethervoid: I'd make some clean in the class like remove commented code like this

ethervoid:
Applying the patch, the command works so good for me :)

app/models/formats/ogr/shp.js
```diff
@@ -63,7 +63,7 @@ ShpFormat.prototype.toSHP = function (options, callback) {
...
         }
         next(null);
```
dgaubert: OMG! This is an anti-pattern, `next` callback is actually called twice. Fixing it...

app/models/formats/ogr/shp.js
dgaubert: 👍

dgaubert:
I've seen a potential bug, please @ethervoid, could review it again?

PR #420, opened 2017-06-08
Support special float values (NaN and Infinity)

dgaubert:
Fixes https://github.com/CartoDB/cartodb/issues/12154 partially

### DIFF
--- a/app/models/formats/pg/json.js
+++ b/app/models/formats/pg/json.js
@@ -66,7 +66,21 @@ JsonFormat.prototype.handleQueryRow = function(row, result) {
 
     this.lastKnownResult = result;
 
-    this.buffer += (this.total_rows++ ? ',' : '') + JSON.stringify(row);
+    this.buffer += (this.total_rows++ ? ',' : '') + JSON.stringify(row, function (key, value) {
+        if (value !== value) {
+            return 'NaN';
+        }
+
+        if (value === Infinity) {
+            return 'Infinity';
+        }
+
+        if (value === -Infinity) {
+            return '-Infinity';
+        }
+
+        return value;
+    });
 
     if (this.total_rows % (this.opts.bufferedRows || 1000)) {
         this.opts.sink.write(this.buffer);
--- a/test/acceptance/query-float-values.js
+++ b/test/acceptance/query-float-values.js
@@ -33,7 +33,10 @@ describe('special numeric (float) values', function() {
                 var next = this;
                 var opts = {
                     url: URL + querystring.stringify({
-                        q: 'insert into numbers_test values (\'NaN\'::float), (\'infinity\'::float), (1::float)'
+                        q: [
+                            'insert into numbers_test',
+                            ' values (\'NaN\'::float), (\'infinity\'::float), (\'-infinity\'::float), (1::float)'
+                        ].join('')
                     }),
                     headers: HEADERS,
                     method: METHOD
@@ -56,9 +59,10 @@ describe('special numeric (float) values', function() {
                 assert.ifError(err);
                 var result = JSON.parse(res.body);
                 assert.ok(Array.isArray(result.rows));
-                result.rows.forEach(function (row) {
-                    assert.ok(row.val !== null, 'special float value (Infinity or NaN) parsed to null');
-                });
+                assert.equal(result.rows[0].val, 'NaN');
+                assert.equal(result.rows[1].val, 'Infinity');
+                assert.equal(result.rows[2].val, '-Infinity');
+                assert.equal(result.rows[3].val, 1);
                 done();
             }
         );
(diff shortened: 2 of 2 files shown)


ethervoid:
I don't think we should allow this behavior for the following reasons:
- We're in `formats/pg/json.js` which, if I'm not wrong, comes from Postgresql where we haven't `Infinity` or `NaN` for number columns.
- Here we're dealing with bad generated data, so convert NaN or Infiity values to NULL looks good to me.
- Returning strings for a numeric column is mixing two kind of data because in the end we're going to transform to JSON again an we're going to have data like this:
```json
{
   "rows":[
      {
         "cartodb_id":1,
         "the_geom":"{\"type\":\"Point\",\"coordinates\":[-3.7038,40.4168]}",
         "latitude":40.4168,
         "longitude":-3.7038,
         "val":1.2
      },
      {
         "cartodb_id":2,
         "the_geom":"{\"type\":\"Point\",\"coordinates\":[-3.7048,40.4178]}",
         "latitude":40.4178,
         "longitude":-3.7048,
         "val":"Infinity"
      },
      {
         "cartodb_id":3,
         "the_geom":"{\"type\":\"Point\",\"coordinates\":[-3.7058,40.4188]}",
         "latitude":40.4188,
         "longitude":-3.7058,
         "val":"NaN"
      }
   ],
   "time":0.002,
   "fields":{
      "cartodb_id":{
         "type":"number"
      },
      "the_geom":{
         "type":"string"
      },
      "latitude":{
         "type":"number"
      },
      "longitude":{
         "type":"number"
      },
      "val":{
         "type":"number"
      }
   },
   "total_rows":3
}
```
- We're relaying in JS only types and from my point of view the response should be agnostic. For example NULL is a wide accepted value for this kind of situations

dgaubert:
I can agree with you from our point of view (backenders 😄 ), we use JSON as data-interchange format which it has a lot of benefits in 99% of the uses cases. 

However, Builder is becoming more and more in a specialised data visualisation tool for data-scientist, analytics and so on. For them is pretty common deal with `infinity` values. As it happened to @andy-esch, some of them are aware that their datasets might have `Infinity` values but when they go to the `dataset view` in Builder they see `null` instead of `Infinity` or `NaN` which is kind of weird for those users.

ethervoid:
Hum didn't read that PostgreSQL allows having NaN, Infinity and -Infinity for floating-point columns 😅

dgaubert:
> We're in formats/pg/json.js which, if I'm not wrong, comes from Postgresql where we haven't Infinity or NaN for number columns.

I don't think so. See https://www.postgresql.org/docs/9.1/static/datatype-numeric.html.

ethervoid:
Yes I've removed that part because is not true :D

dgaubert:
> Here we're dealing with bad generated data, so convert NaN or Infiity values to NULL looks good to me.

Not sure, there will be some use cases that `Infinity` and `NaN` are worth to highlight. So casting them to `null` could lead to misunderstanding.

> Returning strings for a numeric column is mixing two kind of data because in the end we're going to transform to JSON again an we're going to have data like this

Sure, I don't know how builder handles SQL-API responses but I don't think they check if value type correlates to column type. It might someone for Builder could shed some light here.

app/models/formats/pg/json.js
```diff
@@ -66,7 +66,21 @@ JsonFormat.prototype.handleQueryRow = function(row, result) {
 
     this.lastKnownResult = result;
 
-    this.buffer += (this.total_rows++ ? ',' : '') + JSON.stringify(row);
+    this.buffer += (this.total_rows++ ? ',' : '') + JSON.stringify(row, function (key, value) {
```
ethervoid: Let's refactor this, please. Maybe we should change to a proper buffer instead of keeping dealing with this code.

rochoa: @ethervoid Can you outline a bit your proposal?

ethervoid: Yeah sure! I'm not an expert, the comment looks a bit rude and I'm probably very wrong but I'd find this piece of code really hard to read and to manipulate.

I'd simply it, maybe doing it more verbose but better to understand and change.

For example `this.total_rows++`  to check if have to add a `,` or the way we're creating a buffer concatenating strings maybe is not the best (again I should read the internals)

Maybe create an array with the JSON objects, check if we've reached the limit of the buffer, concatenate, concatenate, sink and `.length=0`

I'm pretty sure this code is done this way for performance reasons and I'm being a bigmouth

dgaubert:
Acceptance was performed successfully by @jgoizueta 

@ethervoid for this time I'm going to merge this PR despite your proposal of refactoring since it's a bug and would like to deploy it in production ASAP. 

💋 

Merging...

PR #425, opened 2017-08-02
removed incorrect info from versions page of docs

csobier:
Related to Docs issue: https://github.com/CartoDB/docs/issues/1219

@rochoa , the "Releases" section of this repo does not seem to be related to "version numbers" so I edited the content,  as reported by Support.

Please review and let me know if I can merge until we figure out how/where these "version numbers" are being maintained.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,8 @@
 #Changelog
 
+## 02 August 2017
+Content edits to doc/version.md
+
 ## 1.46.2
 Released 2017-mm-dd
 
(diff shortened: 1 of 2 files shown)


rochoa:
Nice catch!

Do you think we could be more specific about this version being related exclusively to the SQL API?

> They should all contain **/v2/** in the URLs as follows, `https://{username}.carto.com/api/v2/sql`.

csobier:
Yes! i was hoping you would clarify that part. Updating and merging accordingly, thank you!

rochoa:
Thank you. :shipit:

PR #437, opened 2017-10-02
Update README.md

matallo:
/cc @javitonino @ethervoid


dgaubert:
We don't support yarn anymore

PR #446, opened 2017-11-14
Test: Strip PARALLEL labels for PostgreSQL releases before 9.6

Algunenano:
Same patch as done for https://github.com/CartoDB/Windshaft-cartodb/pull/791. Since it's downloading individual files instead of the extension itself it needs to remove PARALLEL labels.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,23 +5,45 @@ addons:
     - ubuntu-toolchain-r-test
     packages:
     - g++-4.8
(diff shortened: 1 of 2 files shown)


Algunenano:
I had to change the travis config a bit because some of the postgresql releases where not being pruned correctly (and a different one would show in pg_config). Also, change the `sed -i` for `mktemp` for OSX compatibility.

PR #448, opened 2017-11-17
Python timeout

oleurud:
When our database timeout limits are exceeded we usually get a proper error message, but if the timeout is hit while the database is executing python code, the error message and status code are not correct.

```
{"error":["RuntimeError: Execution of function interrupted by signal"]}
```

Thanks @jgoizueta for the alert


oleurud:
Looking the problem with the tests in Travis again, I have found something. Before running the tests, after the system configuration, we run this command:
`psql -A -t -c 'select version()'`

The result should be
`PostgreSQL 9.5.2 on ...`

but instead of it, we have:
`PostgreSQL 9.5.7 on ...`

So Travis is running a different version which has not the plpython fix

oleurud:
It is blocked by https://github.com/CartoDB/CartoDB-SQL-API/pull/455

app/postgresql/error_handler.js
```diff
@@ -16,7 +16,10 @@ ErrorHandler.prototype.getMessage = function() {
     // `57014: query_canceled` includes other queries than `statement timeout`, otherwise we could do something more
     // straightforward like:
     // return conditionToMessage[this.err.code] || this.err.message;
-    if (message && message.match(/statement timeout/)) {
+    if (message && 
+        (message.match(/statement timeout/) ||
```
dgaubert: I'd extract `(message.match(/statement timeout/) || message.indexOf('RuntimeError: Execution of function interrupted by signal') > -1)` condition to a function properly named.

yarn.lock
```diff
@@ -0,0 +1,2272 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
dgaubert: Remove this file. This project doesn't require yarn to be installed.

app/postgresql/error_handler.js
dgaubert: In fact, the whole condition. ;)

yarn.lock
oleurud: Thanks :fearful:

oleurud:
[Python timeout error](https://trello.com/c/Scz2Svn9/2-python-timeout-error)

PR #450, opened 2017-11-24
New Auth API

dgaubert:
Hey @enekid, a promise is a promise:

This is functional PR that implements the new Auth API based on scoped permissions. It's not finished yet but it's working and completely backward compatible. 

Pending stuff:
- [x] Tests with scoped tokens.
- ~Do not query metadata in order to get current user's api_key (a.k.a. _pata negra_). It will break backward compatibility, so be careful with that.~


dgaubert:
Oopps!

 I forgot to comment, this PR requires https://github.com/CartoDB/node-cartodb-redis/tree/new-auth-get-api-key-changes to work as expected.

app/controllers/job_controller.js
```diff
@@ -1,146 +1,235 @@
...
 
-        res.header('X-Served-By-DB-Host', req.context.userDatabase.host);
+        next();
+    };
+}
 
-        req.profiler.done(action);
-        req.profiler.end();
-        req.profiler.sendStats();
+const ONE_KILOBYTE_IN_BYTES = 1024;
```
oleurud: Only a comment: I like to show the constants at the beginning and exports at the end (or also beginning) of the file

app/controllers/job_controller.js
```diff
@@ -1,146 +1,235 @@
...
 JobController.prototype.route = function (app) {
+    const { base_url } = global.settings;
+    const forceToBeAuthenticated = true;
+
     app.post(
-        global.settings.base_url + '/sql/job',
-        bodyPayloadSizeMiddleware, userMiddleware, authenticatedMiddleware(this.userDatabaseService),
-        this.createJob.bind(this)
+        `${base_url}/sql/job`,
+        initializeProfilerMiddleware('job'),
```
oleurud: Do we need the profiler before the authentication?

app/controllers/job_controller.js
```diff
@@ -1,146 +1,235 @@
...
+        checkBodyPayloadSize(),
+        userMiddleware(),
+        authorizationMiddleware(this.userDatabaseService, forceToBeAuthenticated),
+        createJob(this.jobService),
+        setServedByDBHostHeader(),
+        finishProfilerMiddleware(),
+        logJobResult('create'),
+        incrementSuccessMetrics(this.statsdClient),
+        incrementErrorMetrics(this.statsdClient),
+        errorMiddleware()
```
oleurud: So nice change!! :grinning: :top:  

Now it is easier to handle errors!

app/middlewares/authorization.js
```diff
@@ -0,0 +1,119 @@
...
+        userDatabaseService.getConnectionParams(authApi, user, function (err, userDbParams, authDbParams, userLimits) {
+            if (req.profiler) {
+                req.profiler.done('setDBAuth');
+            }
+
+            if (err) {
+                return next(err);
+            }
+
+            if (forceToBeAuthenticated && !userDbParams.authenticated) {
```
oleurud: For me, it is weird to see this `userDbParams.authenticated` property after something called `getConnectionParams` instead of something like `checkAuth`. 

Maybe it is something personal, but I have needed to read 2 times to show where the authentication was happening.

dgaubert: > I have needed to read it 2 times to show where the authentication was happening.

Agree, this name comes from the old auth-system which was doing both things: get database credentials and authenticate.

I'm going to tackle this ASAP

app/middlewares/timeout-limits.js
```diff
@@ -0,0 +1,23 @@
+module.exports = function timeoutLimits (metadataBackend) {
```
oleurud: :clap:

app/services/user_database_service.js
```diff
@@ -1,12 +1,20 @@
-'use strict';
+const _ = require('underscore');
 
-var step = require('step');
```
oleurud: :es:

dgaubert:
@enekid 🙏 a CR from your side would be much appreciated!

PR #455, opened 2018-01-08
fix postgres version on travis

oleurud:
Working on https://github.com/CartoDB/CartoDB-SQL-API/pull/448#issuecomment-349339179 I have shown the `postgresql-9.5=9.5.2-3cdb3` and `postgresql-plpython-9.5=9.5.2-3cdb3` are overwritten by `postgresql-9.5-postgis-scripts` installing the default version (currently 9.5.7).
  
As a result, when we run the command `psql -A -t -c 'select version()` before running the tests, we have `PostgreSQL 9.5.7 on ...` instead of the expected `PostgreSQL 9.5.2 on ...`, so we are not working with our custom debian packages

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -31,10 +31,10 @@ before_install:
   - sudo apt-get -y remove --purge postgis-2.2
   - sudo apt-get -y autoremove
 
-  - sudo apt-get -y install postgresql-9.5=9.5.2-3cdb3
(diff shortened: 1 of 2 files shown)


oleurud:
Sure, we must update Travis to use PostGIS 2.3 or 2.4 instead of wasting time on this task, but I need to fix the problem to finish https://github.com/CartoDB/CartoDB-SQL-API/pull/448

Algunenano:
:flags:

Issue #457, opened 2018-02-06
Redis v4 support


oleurud:
Analysis and implementation of Redis v4 support, following the work made in https://github.com/CartoDB/Windshaft-cartodb/issues/819

PR #458, opened 2018-02-08
Redis4 support

oleurud:
From https://github.com/CartoDB/CartoDB-SQL-API/issues/457


oleurud:
@dgaubert @enekid after updating the libraries, I have seen this warning:

```
node_redis: Deprecated: The HMSET command contains a argument of type Object.
This is converted to "[object Object]" by using .toString() now and will return an error from v.3.0 on.
```

You can see in the test we were using `job` instead of `job.data`. Basically, the test it is checking we can't create a job without `job_id` property. 

I have added the code and updated the test to ensure this behavior, but maybe this PR is not the place to add it. Can you review it? ---> https://github.com/CartoDB/CartoDB-SQL-API/pull/458/commits/c3513600855a9b98800c5381bceb251f2c5a618f

dgaubert:
Good catch!

I don't mind if you include this fix in this PR. I prefer to have it fixed already.

PR #469, opened 2018-03-13
Honor `batch_api_port` for Batch Queries

dgaubert:
With the new `Auth API`, a job created with a specific `api-key` should be performed with the same credentials. Before Auth API, all jobs were executed with `master api-key` and used a custom port to bypass `pgbouncer`. After deploying the new Auth API, jobs store the connection params of the user and were saved with the generic `port` instead of the specific one for Batch Queries.

### DIFF
--- a/app/controllers/job_controller.js
+++ b/app/controllers/job_controller.js
@@ -99,7 +99,7 @@ function createJob (jobService) {
             user: res.locals.user,
             query: params.query,
             host: res.locals.userDbParams.host,
-            port: res.locals.userDbParams.port,
+            port: global.settings.db_batch_port || global.settings.db_port,
             pass: res.locals.userDbParams.pass,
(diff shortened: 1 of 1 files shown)


app/controllers/job_controller.js
```diff
@@ -99,7 +99,7 @@ function createJob (jobService) {
             user: res.locals.user,
             query: params.query,
             host: res.locals.userDbParams.host,
-            port: res.locals.userDbParams.port,
+            port: global.settings.db_batch_port || global.settings.db_port,
```
rochoa: Should we do?

```js
port: global.settings.db_batch_port || res.locals.userDbParams.port,
```

dgaubert: 🤔  sure!

dgaubert:
Checking it in staging

dgaubert:
Working! 💪

PR #470, opened 2018-03-23
updates descriptions for python support

andy-esch:


andy-esch:
Besides this change, we should probably not use the 'Fork it' language since people generally just want to use the software (not develop it). Instead, we should say either `install with ...`, or `see the repo for more information`.

PR #471, opened 2018-03-23
Rate limits details

oleurud:
- Remove Rety-after HTTP header when you are not rate limited (now we are returning -1)
- Unify error message with the timeout limit message


app/middlewares/rate-limit.js
```diff
@@ -29,12 +29,16 @@ function rateLimit(userLimits, endpointGroup = null) {
...
                 'Carto-Rate-Limit-Reset': reset
             });
     
             if (isBlocked) {
-                const rateLimitError = new Error('You are over the limits.');
+                // retry is floor rounded in seconds by redis-cell
+                res.set('Retry-After', retry + 1);
+                
+                const rateLimitError = new Error(
+                    'You are over platform\'s limits. Please contact us to know more details'
```
enekid: Instead of using `'... platform\'s ...'` I would instead use `"... platform's ..."` or even better ``` `... platform's ...` ```

PR #472, opened 2018-03-23
timeout error simplify

oleurud:
create a timeout error allowing to add it more properties easily


app/models/formats/ogr.js
```diff
@@ -194,7 +194,7 @@ OgrFormat.prototype.toOGR = function(options, out_format, out_filename, callback
         clearTimeout(ogrTimeout);
 
         if (timedOut) {
-          return next(new Error('You are over platform\'s limits. Please contact us to know more details'));
+          return next(new Error('statement timeout'));
```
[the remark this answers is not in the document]
oleurud: If we don't change it, the error won't be cached by ErrorHandler constructor.

Before these changes, we were creating the timeout limit error in 2 different places, now, my idea is to create only in the ErrorHandler constructor. So, to make it possible, I need to change the ogr error like a Postgres error.

Another option is to use another string to manage the error instead of using a postgres one.

app/postgresql/error_handler.js
```diff
@@ -11,18 +18,11 @@ ErrorHandler.prototype.getName = function() {
 };
 
 ErrorHandler.prototype.getMessage = function() {
-    var message = this.err.message;
-
-    if (this.isTimeoutError()) {
```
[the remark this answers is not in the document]
oleurud: Because the timeout error is caught on the constructor and created a new one with all the parameters: message and HTTP status code. So, we don't need to make special things on the rest of the object ErrorHandler for timeout error

app/models/formats/ogr.js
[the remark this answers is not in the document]
dgaubert: got it

app/postgresql/error_handler.js
```diff
@@ -11,18 +18,11 @@ ErrorHandler.prototype.getName = function() {
...
-
-    return message;
+    return this.err.message;
 };
 
-ErrorHandler.prototype.getFields = function(fields) {
-    fields = fields || ['detail', 'hint', 'context'];
-    return fields.reduce(function (previousValue, current) {
+ErrorHandler.prototype.getFields = function() {
+    return ['detail', 'hint', 'context'].reduce(function (previousValue, current) {
```
enekid: I would refactor this to an arrow function.

```javascript
return ['detail', 'hint', 'context'].reduce((fields, fieldName) => {
    fields[fieldName] = this.err[fieldName];
    return fields;
}, {});
```

And probably I would use a `forEach`:


```javascript
const fields = {};
['detail', 'hint', 'context'].forEach( fieldName => fields[fieldName] = this.err[fieldName])
return fields;  
```

But that's up to your taste.

test/acceptance/app.test.js
```diff
@@ -792,7 +792,9 @@ it('GET with callback must return 200 status error even if it is an error', func
                 assert.deepEqual(error, {
                     error: [
                         'You are over platform\'s limits. Please contact us to know more details'
-                    ]
+                    ],
+                    context: "limit",
```
enekid: Eneko linter speaking here:
Prefer `'...'` over `"..."`

app/postgresql/error_handler.js
oleurud: I have made a refactor of this method (too much complexity unneeded) but I didn't commit it to not mix tasks, but I like your proposal because it is more legible. What do you think @dgaubert ?

oleurud: My approach was:
```
return {
        detail: this.err.detail,
        hint: this.err.hint,
        context: this.err.context,
    };
```

enekid: I like your approach. Much simpler.

Issue #474, opened 2018-03-27
Error middleware


oleurud:
The objective of this PR is to break down the error middleware making it more modular using several middlewares and following the single responsibility principle.

Initial objectives:
- create error class
- differentiate errors from Postgres and the rest
- break in small modules
- ErrorHandler refactor

oleurud:
Deploy process:
- Staging cloud
- 1 server in production
- Merge
- Master deploy

PR #475, opened 2018-03-28
Error middleware refactor

oleurud:
Solves: https://github.com/CartoDB/CartoDB-SQL-API/issues/474

Main changes:
- ErrorHandler changes: now it is a generic error class that extends the Error class with our error properties: `http_status`, `context`, `detail`, and `hint`
- creation of errorHandlerFactory: receives an error (postgres error, timeout error, rate limit error, ...) and returns an ErrorHandler instance error.

Weird changes:
- Removing ErrorHandler dependency from format models: json and geojson. They used ErrorHandler passing an error message and then using the message from ErrorHandler. I have changed it using the error message directly. It wasn't an objective, but I needed to change it to make the rest.


app/middlewares/error.js
```diff
@@ -1,41 +1,35 @@
...
-    return function errorMiddleware (err, req, res, next) {
-        var pgErrorHandler = new PgErrorHandler(err);
-
-        var msg = {
-            error: [pgErrorHandler.getMessage()]
-        };
-
-        _.defaults(msg, pgErrorHandler.getFields());
+module.exports = function errorMiddleware() {
+    return function error(err, req, res, next) {
```
enekid: We usually name these functions this way:
```javascript
module.exports = function error() {
  return function errorMiddleware(err, req, res, next) {
```

app/services/error_handler.js
```diff
@@ -0,0 +1,34 @@
+class ErrorHandler extends Error {
+    constructor(message, context, detail, hint, http_status = 400, name = null) {
```
enekid: I would use an object instead of this parameter list:

```javascript
constructor({message, context, detail, hint, http_status = 400, name = null})
```

This way, calls to the constructor will be more understandable:

```javascript
return new ErrorHandler(
        message: 'You are over platform\'s limits. Please contact us to know more details',
        context: 'limit',
        detail: 'datasource',
        hint: undefined,
        http_stauts: 429
    );
```

app/services/error_handler.js
```diff
@@ -0,0 +1,34 @@
...
+        return {
+            error: [this.message],
+            context: this.context,
+            detail: this.detail,
+            hint: this.hint
+        };
+    }
+    
+    getHttpStatus (http_status) {
+        if (this.message.includes('permission denied')) {
```
dgaubert: :)

app/middlewares/error.js
dgaubert: Yep, please be consistent here.

oleurud:
How do you see it @enekid ?

PR #478, opened 2018-04-23
Postgres entities access validator

oleurud:
Fix: https://github.com/CartoDB/cartodb-management/issues/5157


app/services/pg-entities-access-validator.js
```diff
@@ -0,0 +1,60 @@
...
+            ) {
+                return false;
+            }
+        }
+
+        return true;
+    },
+
+    softValidation(tables) {
+        for (let table of tables) {
```
enekid: I would love to use lodash in these scenarios:

```javascript
const _ = require('lodash');
softValidation(tables) {
  return ! _.some(tables, table => table.table_name.match(/\bpg_/);
}
```

app/services/pg-entities-access-validator.js
```diff
@@ -0,0 +1,60 @@
...
+};
+
+
+module.exports = {
+    validate(affectedTables, authenticated) {
+        let hardValidationResult = true;
+        let softValidationResult = true;
+
+        if (!!affectedTables && affectedTables.tables) {
+            if (global.pgEntitiesAccessValidator) {
```
enekid: rename this flag to something like: ```validatePGEntitiesAccess```

app/services/pg-entities-access-validator.js
```diff
@@ -0,0 +1,60 @@
...
+    information_schema: ['*'],
+    public: ['spatial_ref_sys'],
+    topology: [
+        'layer',
+        'topology'
+    ]
+};
+
+
+module.exports = {
```
enekid: Do we need to export all these 3 functions? Isn't it enough with `validate`?

app/services/pg-entities-access-validator.js
oleurud: I would prefer not to use external dependencies if the result and performance is the same

oleurud: And it is easy to understand ;)

app/services/pg-entities-access-validator.js
oleurud: :+1:

app/services/pg-entities-access-validator.js
oleurud: I want to add some unit test for these functions

config/environments/development.js.example
```diff
@@ -120,3 +120,5 @@ module.exports.ratelimits = {
     job_delete: false
   }
 }
+
+module.exports.validatePGEntitiesAccess = false;
```
enekid: This should be documented *Somewhere*.

app/services/pg-entities-access-validator.js
```diff
@@ -0,0 +1,61 @@
...
+    pg_catalog: ['*'],
+    information_schema: ['*'],
+    public: ['spatial_ref_sys'],
+    topology: [
+        'layer',
+        'topology'
+    ]
+};
+
+const Validator = {
```
dgaubert: Exporting a set of methods in this way only makes sense if you are going to use it as a prototype of a function constructor. Otherwise, I'd go for the well known `module.exports. validate = function ...`.

PR #480, opened 2018-04-26
/copyfrom and /copyto end points

pramsey:
**NOTE: Still a work-in-progress, PR created for purposes of review and feedback.**

Adds support for streaming data in/out of the database using the `COPY` command and new end points.

### `/copyfrom`

Reads a file POST'ed to the end point and streams the contents into a PostgreSQL [COPY ... FROM](https://www.postgresql.org/docs/current/static/sql-copy.html).

* POST to `/copyfrom` only
* `q` parameter contains the [COPY](https://www.postgresql.org/docs/current/static/sql-copy.html) command, in the URL 
* POST body mime-type `application/octet-stream`
* POST body contains the COPY file, in the format described by the COPY command
* Example: 
  ```curl -X POST -H "Content-Encoding: gzip" -H "Transfer-Encoding: chunked" -H "Content-Type: application/octet-stream" --data-binary @upload_example.csv.gz "http://cdb.localhost.lan:8080/api/v2/sql/copyfrom?api_key={api_key}&q=COPY+upload_example+(the_geom,+name,+age)+FROM+STDIN+WITH+(FORMAT+csv,+HEADER+true)"```
* Returns: `{"time":0.061, "total_rows":3}`

### `/copyto`

Runs a PostgreSQL [COPY ... TO](https://www.postgresql.org/docs/current/static/sql-copy.html) and streams the result to the caller as an `application/octet-stream` download.

* GET from `/copyto` only
* response mime-type `application/octet-stream`
* `filename` parameter contains the desired file name for the `Content-disposition` header
* `sql` parameter contains the `COPY ... TO` command to run
* example: `curl "http://cdb.localhost.lan:8080/api/v2/copyto?sql=copy+foo+to+stdout&filename=output.dmp"`


pramsey:
Associated infra ticket: https://github.com/CartoDB/cartodb-platform/issues/4369

pramsey:
@oleurud / @ethervoid could I have a code review? cc/ @rafatower

rafatower:
Related: https://github.com/CartoDB/CartoDB-SQL-API/pull/485

:wave: reviewers

dgaubert:
See: https://github.com/CartoDB/CartoDB-SQL-API/pull/485

pramsey:
Note to @oleurud, @rafatower, @dgaubert and @pramsey, this is now the /copy branch, all the review branches are merged and closed.

app/middlewares/body-parser.js
```diff
@@ -141,5 +141,26 @@ exports.parse['application/json'] = function(req, options, fn){
     });
 };
 
-var multipartMiddleware = multer({ limits: { fieldSize: Infinity } });
-exports.parse['multipart/form-data'] = multipartMiddleware.none();
+
+exports.parse['multipart/form-data'] = function(req, options, fn){
```
oleurud: As we finally don't use the multipart form, in order to have the things small, I would like to revert this change, returning to `multer` instead of `busboy`.

doc/copy_queries.md
```diff
@@ -0,0 +1,184 @@
+# Copy Queries
```
oleurud: Awesome documentation :heart:

app/middlewares/body-parser.js
rafatower: @dgaubert any opinions? whatever you decide works for me, just trying to keep the wheels turning :)

oleurud: I have reverted it. Sincerely, I prefer the busboy solution, because you have more control over the request, but this PR is no longer the place

oleurud:
For me the code looks good. 

We already have task (limits, auth) to define before merge it, am I right?

app/middlewares/body-parser.js
dgaubert: @oleurud is right. Since we are not using `multipart/form-data` anymore the change of replacing `busboy` by `multer` has no sense. So keeping `multer` in this PR is fine.

oleurud:
Should we add gzip support for `copyto` endpoint? 
- Pros: obviously faster download
- Cons: high CPU usage in order to gzip the response

rafatower:
Hey, can  this review be considered OK? can we proceed to [staging](https://github.com/CartoDB/cartodb-platform/issues/4369#issuecomment-389764321)?  thanks!

oleurud:
I am going to create a new issue related to this task, in order to think the best way to connect with Postgres for the current case (pgbouncer?, pg pool?, both? any?), and also to add any related comment to it (better to use an issue instead of a PR).

And at the same time, and following the @rafatower comment  https://github.com/CartoDB/product/issues/60#issuecomment-389482885, I am going to deploy it to staging in the current state, in order to check the global viability of the feature

pramsey:
100% we should stream the `copyto` responses with `Content-encoding: gzip`. These are huge text files, and compression is absolutely required to get the performance improvements we want. Now, whether we need to implement content encoding ourselves or whether we get it for free from nginx is a question I leave for bigger brains.

enekid:
Compression/Decompression should **always** be done by NGINX. Node.js is not particularly good handling this type of CPU intensive tasks (it would block the event loop). The same happens with all the SSL stuff.

Configuring NGINX for handling compression is fairly [trivial](https://docs.nginx.com/nginx/admin-guide/web-server/compression/)

andrewbt:
suggesting to use the "q" parameter instead of "sql" per the rest of the SQL API: https://carto.com/developers/sql-api/reference/#operation/getSQLStatement

rafatower:
@enekid et al.

> Compression/Decompression should always be done by NGINX. Node.js is not particularly good handling this type of CPU intensive tasks (it would block the event loop). The same happens with all the SSL stuff.

Compressing a huge file in a batch is clearly a CPU-bound operation. But... can you prove that the whole thing (streaming end to end) is CPU-bound before deciding node is not suited to the task?

pramsey:
Developer's center pull request for swagger API docs is https://github.com/CartoDB/developers/pull/338

app/controllers/copy_controller.js
```diff
@@ -0,0 +1,194 @@
...
+            const pg = new PSQL(res.locals.userDbParams);
+            pg.connect(function (err, client) {
+                if (err) {
+                    return next(err);
+                }
+
+                const copyToStream = copyTo(sql);
+                const pgstream = client.query(copyToStream);
+                pgstream
+                    .on('error', next)
```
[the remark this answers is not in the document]
oleurud: The result of
```
.on('error', (err) => {
  next(err);
})
```
and
```
.on('error', next)
```
is the same.

And about the catch: we need to change/fix some things. 
- Probably a big try/catch it is not a good option (I don't change it because I need to look if something can throw). 
- How we handle the db connection

app/services/stream_copy.js
```diff
@@ -61,8 +61,9 @@ module.exports = {
                 })
                 .on('close', () => {
                     if (!requestEnded) {
+                        const connection = client.connection;
+                        connection.sendCopyFail();
```
rafatower: :clap: :clap: :clap: :clap:

app/services/stream_copy.js
```diff
@@ -0,0 +1,132 @@
...
+                .on('end', () => {
+                    metrics.end(copyToStream.rowCount);
+                    done();
+                    return cb(null, metrics);
+                })
+                .pipe(res);
+        });
+    },
+
+    from (req, sql, userDbParams, user, gzip, logger, cb) {
```
dgaubert: I think that both `to` and `from` methods are doing a lot of things (one smell is that they receive many parameters). They encapsulate almost everything from the middleware, which is fine but I think that some things are better to handle in the middleware than here. Middlewares should be the ones responsible for `req` and `res` and we shouldn't pass them to collaborators.

If this method returns the pgcopy stream ready to be piped to the request, something like this:

```javascript
class StreamCopy {
  contructor (...) {
    ...
  }

  from (userDbParams, metrics, callback) {
    [ ... ]
    const pg = new PSQL(userDbParams);
    pg.connect(function (err, client, done) {
      [ ... ]
      const copyFromStream = copyFrom(sql);
      const pgstream = client.query(copyFromStream);

      pgstream.on(...);
      pgstream.on(...);

      callback(null, pgstream);
    });
  }
}
```

And then, the middleware would look like: 
```javascript
function handleCopyFrom (logger) {
    return function handleCopyFromMiddleware (req, res, next) {
        [...]
        streamCopy.from(userDbParams, metrics, function(err, pgstream) { 
            if (err) {
                return next(err);
            } 

            let requestEnded = false;
            req.on('error', err => { [...] })
            req.on('close', () => { [...] })
            req.on('data', data => { [...] })
            req.on('end', () => requestEnded = true);
            
            [...]
            res.send({ time, total_rows: rows});
        });
    };
}
```

Obviously, I'm missing some stuff here, it just an example, the middleware is doing a lot of stuff (everything related to `req` and `res`) and will need to be refactored.

What do you think?

oleurud: I already did a refactor splitting controller, stream-copy and metrics. But I agree, stream-copy is managing 2 responsibilities yet, and also, it has grown.

But it is not an easy change, especially on the copy to case. I am going to try it

oleurud: I have created a new branch with the requested changes: https://github.com/CartoDB/CartoDB-SQL-API/pull/503/files

rafatower:
As per our conversation, I'm merging this PR.

@oleurud could you please create a ticket for the intended refactor?

alasarr:
👏👏 That's true love! Best PR ever 😍

PR #481, opened 2018-04-26
fix validatePGEntitiesAccess access

oleurud:
I added an error in https://github.com/CartoDB/CartoDB-SQL-API/pull/478

As a result, the new validatePGEntitiesAccess is never used (and always works as `false`)

### DIFF
--- a/app/services/pg-entities-access-validator.js
+++ b/app/services/pg-entities-access-validator.js
@@ -20,7 +20,7 @@ const Validator = {
         let softValidationResult = true;
 
         if (!!affectedTables && affectedTables.tables) {
-            if (global.validatePGEntitiesAccess) {
+            if (global.settings.validatePGEntitiesAccess) {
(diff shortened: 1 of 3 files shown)


app/services/pg-entities-access-validator.js
```diff
@@ -20,7 +20,7 @@ const Validator = {
         let softValidationResult = true;
 
         if (!!affectedTables && affectedTables.tables) {
-            if (global.validatePGEntitiesAccess) {
+            if (global.settings.validatePGEntitiesAccess) {
```
dgaubert: If we don't define this in the configuration it's going to work anyway because `undefined` is a `falsy` value.  However, I'd try to add a default value in case that `validatePGEntitiesAccess` is `undefined` in order to avoid possible breaking changes like https://github.com/CartoDB/CartoDB-SQL-API/issues/484

PR #485, opened 2018-05-03
Pgcopy stream review

oleurud:
- [x] Adding tests
- [x] New rate limits endpoints
- [x] Solve bodyParser problem
- [x] Minor ES6 changes


app/controllers/copy_controller.js
```diff
@@ -76,8 +76,8 @@ CopyController.prototype.route = function (app) {
         ];
     };
 
-    app.post(`${base_url}/copyfrom`, copyFromMiddlewares(RATE_LIMIT_ENDPOINTS_GROUPS.QUERY));
-    app.get(`${base_url}/copyto`, copyToMiddlewares(RATE_LIMIT_ENDPOINTS_GROUPS.QUERY));
+    app.post(`${base_url}/sql/copyfrom`, copyFromMiddlewares(RATE_LIMIT_ENDPOINTS_GROUPS.COPY_FROM));
```
pramsey: I don't see the COPY_FROM/COPY_TO entries being added to the look-up table in the rate limits side, is that coming?

oleurud: There are in `app/middlewares/rate-limit.js`

We will need to talk with Ruby's people to add the rate limit values of the new endpoints in Redis 

cc/ @javitonino @alrocar

alrocar: @oleurud Can you add an issue to cartodb-platform (to keep it private) and assign it to me please?

Just add the name(s) of the endpoint(s) and the 3 values for the rate limits and I'll take care.

Any deadline for this?

oleurud:
I don't like my solution to `bodyParser problem` moving bodyParser to the routes of query and job controllers.

I had thought of using 2 express Router to solve it: a router to query and job controllers with the current bodyParser and another router to copy controller. But it is not possible due to the routes definition, all of the query and job controllers start with `/sql` but we have `/jobs-wip`...

And I don't see any other solution. If someone has another, ping me

pramsey:
Why don't you like your solution? so much of other policy is per-controller, it actually seemed odd to me when I found it that this one piece of policy was global.

oleurud:
I would prefer not to repeat the code. But it is not bad as is

app/utils/multer-pg-copy.js
```diff
@@ -11,27 +10,17 @@ function PgCopyCustomStorage (opts) {
...
-}
-
 PgCopyCustomStorage.prototype._handleFile = function _handleFile (req, file, cb) {
 
     // Hopefully the body-parser has extracted the 'sql' parameter
     // or the user has provided it on the URL line.
     // Otherwise, this will be a short trip, as we won't be able
     // to the pg-copy-streams SQL command
-    var sql = getSqlParam(req);
+    var sql = req.body.sql || req.query.sql;
```
enekid: IMHO querystring params should have preference over body params.

enekid: If we go with this _sillyness-proven_ way we should probably refactor all our APIs to be consistent.

Let's check what we are currently doing in such cases and go with it.

We should also consider documenting this new feature, adding these type of quirks

app/controllers/copy_controller.js
pramsey: Let's aim for next week, so we can perhaps have everything merged by next week and perhaps get to play in staging the week following? (laughable? I hope not...)

app/utils/multer-pg-copy.js
```diff
@@ -11,39 +10,29 @@ function PgCopyCustomStorage (opts) {
...
 
     // We expect the an earlier middleware to have 
     // set this by the time we are called via multer,
     // so this should never happen
     if (!req.userDbParams) {
-        cb(new Error("req.userDbParams is not set"));
+        return cb(new Error("Something was wrong"));
     }
 
     var copyFromStream = copyFrom(sql);
```
dgaubert: Since we are not storing `req.body` into a file we don't need to remove it in `_removeFile` as `file.path` was never set.

app/controllers/copy_controller.js
oleurud: @alrocar this is the issue  --> https://github.com/CartoDB/cartodb-platform/issues/4394

app/utils/multer-pg-copy.js
oleurud: I have done a reset HEAD deleting the master merge.

We can simplify the process with `busboy`. I am going to try it

dgaubert: And having access to all `busboy`'s options we can tune limits, back-pressure and so on

app/controllers/copy_controller.js
alrocar: Thanks, added to the backend kanban

app/controllers/copy_controller.js
```diff
@@ -72,85 +47,140 @@ CopyController.prototype.route = function (app) {
             connectionParamsMiddleware(this.userDatabaseService),
             timeoutLimitsMiddleware(this.metadataBackend),
             this.handleCopyTo.bind(this),
+            this.responseCopyTo.bind(this),
```
pramsey: I find it kind of weird that the headers are set *after* the stream is set up, but then much of node.js seems weird to me :)

pramsey:
👍

app/controllers/copy_controller.js
```diff
@@ -11,42 +8,22 @@ const timeoutLimitsMiddleware = require('../middlewares/timeout-limits');
 const { initializeProfilerMiddleware } = require('../middlewares/profiler');
 const rateLimitsMiddleware = require('../middlewares/rate-limit');
 const { RATE_LIMIT_ENDPOINTS_GROUPS } = rateLimitsMiddleware;
+const Busboy = require('busboy');
```
dgaubert: It's a bit weird requiring a module that it's actually a sub-dependency of the main dependency (`multer`). Or we define `busboy` as a dependency in `package.json` or we make use of `multer`.

app/middlewares/body-parser.js
```diff
@@ -11,7 +11,7 @@
  */
 
 var qs = require('qs');
-//var multer = require('multer');
+var multer = require('multer');
```
dgaubert: Oh my goodness! I didn't know that we were using `multer` in this middleware since the beginning. This changes totally my suggestion of using `busboy` instead. 

Sorry.

oleurud: I am going to try! No much in code, but not sure if it will be easy to do with busboy

oleurud:
I have changed the multer by busboy in multipart requests. I'm not completely sure, but I'm pretty sure about it.

oleurud:
Before merging it, we want to make an acceptance. Tomorrow I will put this branch on staging server.

pramsey:
As I've picked up the problem of GDAL support for /copy, my decision to use multipart/form-data for the file upload is bugging me (because libcurl support for it is pretty thin). Using multipart/form-data does allow both the sql and file parameters to be in the POST body, which is nice, but it's not required. If the sql parameter was in the URL, then the body could just be the file content, and sent up via a Chunked transfer encoding. Data would still be streaming, and it would probably be easier to stream data for which you do not know the file size _a priori_. (One of the issues in libcurl is that you cannot stream a file in multipart/form-data without knowing its size in advance.) For OGR, being able to stream w/o knowing the size would be handy, since you don't know the size of the eventual output while iterating the input records. cc/ @rafatower

oleurud:
This branch is deployed to staging cloud

rafatower:
Taking into account Paul's [comment](https://github.com/CartoDB/CartoDB-SQL-API/pull/485#issuecomment-388042496) I'd hold the release of this and give a try to chunked encoding because of two main reasons:
- the need of knowing the size _a priori_ can be a limitation not just to ogr2ogr and the underlying libcurl but to other clients.
- this decision affects the API (and also, potentially, how limits are implemented)

cc/ @rochoa

pramsey:
OK, I flipped the implementation to accept a chunked POST instead, in #487, which I guess should be merged here, after some help from @oleurud on getting the tests running again (not sure how to post the file data in the test framework).

Issue #484, opened 2018-05-03
TypeError: Cannot read property 'rateLimitsEnabled' of undefined


rochoa:
The [`ratelimits` configuration key](https://github.com/CartoDB/CartoDB-SQL-API/blob/6ad54b284f11eb401f9e187d525caf6f3ec58a03/app/server.js#L142) is not backwards compatible.

### Current result
Old deployments fail to start if the new key is not defined in an old configuration file.

### Expected behaviour
Be able to use an old configuration file with the new code.

P.S.: The [release notes for 2.x do not cover this change as a breaking one](https://github.com/CartoDB/CartoDB-SQL-API/releases/tag/2.0.0).

oleurud:
I expected the changes on environment configuration files were a must, but of course, I should have warned about it in the release. 

Luckily, the problem doesn't happen in `Windshaft-carto`.

PR #486, opened 2018-05-08
Making ratelimits configuration keys backwards compatible

oleurud:
Fix: https://github.com/CartoDB/CartoDB-SQL-API/issues/484

### DIFF
--- a/app/server.js
+++ b/app/server.js
@@ -139,7 +139,7 @@ function App(statsClient) {
     
     const userLimitsServiceOptions = {
         limits: {
-            rateLimitsEnabled: global.settings.ratelimits.rateLimitsEnabled
+            rateLimitsEnabled: (global.settings.ratelimits && global.settings.ratelimits.rateLimitsEnabled) || false
         }
     };
     const userLimitsService = new UserLimitsService(metadataBackend, userLimitsServiceOptions);
--- a/app/middlewares/rate-limit.js
+++ b/app/middlewares/rate-limit.js
@@ -51,8 +51,10 @@ function rateLimit(userLimits, endpointGroup = null) {
 }
 
 function isRateLimitEnabled(endpointGroup) {
-    return global.settings.ratelimits.rateLimitsEnabled &&
+    return global.settings.ratelimits &&
+        global.settings.ratelimits.rateLimitsEnabled &&
         endpointGroup &&
+        global.settings.ratelimits.endpoints &&
         global.settings.ratelimits.endpoints[endpointGroup];
 }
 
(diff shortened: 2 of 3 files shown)


dgaubert:
Do you need a CR here?

app/server.js
```diff
@@ -139,7 +139,7 @@ function App(statsClient) {
     
     const userLimitsServiceOptions = {
         limits: {
-            rateLimitsEnabled: global.settings.ratelimits.rateLimitsEnabled
+            rateLimitsEnabled: (global.settings.ratelimits && global.settings.ratelimits.rateLimitsEnabled) || false
```
enekid: another way:
```javascript
rateLimitsEnabled: Boolean(global.settings.ratelimits && global.settings.ratelimits.rateLimitsEnabled)
```

oleurud:
I have changed my point of view to solve it. Initially, my idea was adding more controls avoiding the `Cannot read property 'rateLimitsEnabled' of undefined` adding more checks. But I don't like it because the checks (although simples) were in the middle of the rate limit process. 

I have added a configuration method ensuring the correct parameters and It makes the control explicit. I would prefer something like server_options of Windshaft-carto, but it is not bad

app/services/user_limits.js
```diff
@@ -11,6 +22,13 @@ class UserLimits {
         this.preprareRateLimit();
     }
 
+    static configure() {
+        // default rate limits
+        if(!global.settings.ratelimits) {
+            global.settings.ratelimits = DEFAULT_RATE_LIMITS_OPTIONS;
```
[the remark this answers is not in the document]
oleurud: :+1:

app/services/user_limits.js
```diff
@@ -11,6 +22,13 @@ class UserLimits {
         this.preprareRateLimit();
     }
 
+    static configure() {
```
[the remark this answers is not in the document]
oleurud: Initially, I have added it to server.js (we can't add it to app.js because the tests don't use it). But finally, I have preferred to write it in the related code (rate limits are managed in user_limits.js), in order to maintain server.js small. 

But I don't have a preference, I don't really like in any case.

app/server.js
```diff
@@ -66,6 +66,19 @@ function App(statsClient) {
     // Set default configuration
     global.settings.db_pubuser = global.settings.db_pubuser || "publicuser";
     global.settings.bufferedRows = global.settings.bufferedRows || 1000;
+    global.settings.ratelimits = Object.assign(
+        {
+            rateLimitsEnabled: false,
+            endpoints: {
+                query: false,
+                query_format: false,
```
enekid: Just for the sake of pickiness: Why are those properties named with `_` as word separators instead of the standard camelCase way?

oleurud: Because they are used in Ruby and Redis too. We find a consensus between the parts, and this was the result.

PR #487, opened 2018-05-10
Change from form-multipart to POST with chunked upload

pramsey:
For @oleurud to have a look it: the change to POST is minor. The changes to the tests are probably nonsensical: I had to guess at how to do a POST-with-data in the test framework.


oleurud:
@pramsey Your tests were fine for me. I have only changed some details: https://github.com/CartoDB/CartoDB-SQL-API/pull/487/commits/85528459b85fcb5371603427e09ee4bcb34bc651

I not sure about it, but I have thought that you want to add gzip support and I have added it. I have added it only in `copyfrom` due to add it in `copyto` wiil be a CPU intensive task. Anyway, feel free to remove it if you think that it has not sense.

Also, I have changed the headers of `copyto` that I forgot to change in the previous PR.

PR #490, opened 2018-05-21
Pgcopy stream metrics

oleurud:
In order to measure the behavior of the copy endpoints (following https://github.com/CartoDB/cartodb-platform/issues/4414)

Things that are interesting to be measured: 
- query time, 
- size of payload, 
- format (text, csv, or binary)

### DIFF
--- a/app/controllers/copy_controller.js
+++ b/app/controllers/copy_controller.js
@@ -156,14 +156,14 @@ CopyController.prototype.responseCopyFrom = function (req, res, next) {
 
     if (req.profiler) {
         const copyFromMetrics = {
-            time: res.body.time, //seconds
             size: res.locals.copyFromSize, //bytes
+            time: res.body.time, //seconds
             total_rows: res.body.total_rows, 
             gzip: req.get('content-encoding') === 'gzip'
         };
 
         req.profiler.add({ copyFrom: copyFromMetrics });
-        res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());+
+        res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());
     }
 
     res.send(res.body);
--- a/app/controllers/copy_controller.js
+++ b/app/controllers/copy_controller.js
@@ -106,6 +106,8 @@ CopyController.prototype.handleCopyFrom = function (req, res, next) {
         return next(new Error("SQL must start with COPY"));
     }
 
+    res.locals.copyFromSize = 0;
+
     try {
         const start_time = Date.now();
 
@@ -130,9 +132,14 @@ CopyController.prototype.handleCopyFrom = function (req, res, next) {
             });
 
             if (req.get('content-encoding') === 'gzip') {
-                req.pipe(zlib.createGunzip()).pipe(pgstream);
+                req
+                    .pipe(zlib.createGunzip())
+                    .on('data', data => res.locals.copyFromSize += data.length)
+                    .pipe(pgstream);
             } else {
-                req.pipe(pgstream);
+                req
+                    .on('data', data => res.locals.copyFromSize += data.length)
+                    .pipe(pgstream);
             }
         });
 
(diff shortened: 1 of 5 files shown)


app/utils/query_info.js
```diff
@@ -0,0 +1,21 @@
...
+
+        copyQuery = copyQuery.toUpperCase();
+
+        if (!copyQuery.toUpperCase().startsWith("COPY ")) {
+            return false;
+        }
+       
+        const regex = /(\bFORMAT\s+)(\w+)/;
+        const result = regex.exec(copyQuery);
+        if (result && result.length >= 3 && COPY_FORMATS.includes(result[2])) {
```
rafatower: Minor thing: most likely you don't need a `(capture group)` for the `\bFORMAT\s+`. Also, you don't expect more than 1 format in the query string. So, you could write

```js
        const regex = /\bFORMAT\s+(\w+)/;
        const result = regex.exec(copyQuery);
        if (result && result.length === 2 && COPY_FORMATS.includes(result[1])) {
            //...
        }
```

app/utils/query_info.js
```diff
@@ -0,0 +1,21 @@
...
+            format = result[2];
+        }
```
rafatower: The default format, if not present in options is `text`. I guess we want to have it in the stats for simplicity.

test/unit/query_info.test.js
```diff
@@ -0,0 +1,49 @@
...
+            });
+        });
+
+        describe('text', function() {
+            const csvValidQueries = [
+                "COPY copy_endpoints_test (id, name) FROM STDIN WITH (FORMAT TEXT)",
+                "COPY copy_endpoints_test (id, name) FROM STDIN",
+            ];
+
+            csvValidQueries.forEach(query => {
```
rafatower: > csvValidQueries

csv or text ?

app/utils/query_info.js
rafatower: ah, sorry, did not see the default above!

app/utils/query_info.js
oleurud: Nice, thank you!

test/unit/query_info.test.js
oleurud: :scream:  copy paste

oleurud:
Reviewers, take into account ---> https://github.com/CartoDB/cartodb-platform/issues/4414#issuecomment-390932252

app/controllers/copy_controller.js
```diff
@@ -55,99 +57,132 @@ CopyController.prototype.route = function (app) {
...
+            size: 0,
+            time: null,
+            format: getFormatFromCopyQuery(req.query.sql)
+        };
+
+        res.header("Content-Disposition", `attachment; filename=${encodeURIComponent(filename)}`);
+        res.header("Content-Type", "application/octet-stream");
+
+        try {
+            const start_time = Date.now();
```
enekid: `startTime` :)

app/controllers/copy_controller.js
```diff
@@ -55,99 +57,132 @@ CopyController.prototype.route = function (app) {
...
+                    const end_time = Date.now();
+                    res.body = {
+                        time: (end_time - start_time) / 1000,
+                        total_rows: copyFromStream.rowCount
+                    };
+
+                    return next();
+                });
+
+                if (req.get('content-encoding') === 'gzip') {
```
enekid: hopefully we can handle this gunziping with nginx, but at the momment it's ok

app/utils/query_info.js
```diff
@@ -0,0 +1,21 @@
+const COPY_FORMATS = ['TEXT', 'CSV', 'BINARY'];
```
enekid: If the user specifies an invalid format (eg: `WADUS`), we will set the format to `TEXT`. Shouldn't we return an error, or at least a `false`

app/controllers/copy_controller.js
oleurud: :+1:

app/controllers/copy_controller.js
oleurud: https://github.com/CartoDB/cartodb-platform/issues/4447

app/utils/query_info.js
oleurud: If the user sends a bad format, the request will fail. 

But for sure, this module should work independently. I am going to change it

oleurud: I have made some changes... not perfect, but enough for me

PR #491, opened 2018-05-21
Change from 'sql' as query parameter to 'q', aping existing sql api

pramsey:


pramsey:
Also updated the doco, where it was still incorrectly referencing multi-part form uploads.

PR #493, opened 2018-05-28
Psql update

Algunenano:


Algunenano:
Merging for now. Release pending.

PR #495, opened 2018-05-29
[DNM] Cancel copy from

rafatower:
With the right parameters. Based on this:

https://github.com/brianc/node-postgres/pull/1392/files#diff-32a8cfb1af74860f3233cd24dcd57fad

This is a bit tricky, but the way it should be at lower levels (cartodb-psql and pg levels).

cc/ @oleurud


rafatower:
Closing this, as it won't be merged in the current state. The solution for the COPY FROM was implemented in 310f652

PR #497, opened 2018-05-31
Cancel copy to upon client disconnect

rafatower:
Hey, I have this partial solution but it has a flaw: connections are not returned to the pg (node) pool (I'm wondering if they are in the case of the COPY FROM and CopyFail command).

How to test it:
1. Use a direct connection to PG: `module.exports.db_port  = '5432';` in config file.
1. Set `module.exports.db_pool_size = 5` in configs to a rather small number.
1. Perform 5  of `COPY TO` operations and cancel them (curl + `CTRL+C`).
1. Try to complete the 6th COPY TO.

Expected: the 6th operation takes an established connection from the pool and executes it. The COPY TO completes.

Actual: as there are no pg connections available in the pg pool, the 6th operation is enqueued waiting for one to be placed back in the pool, and curl eventually times out.

You can also see the connection/backends in the DB:
```
select pid, state, wait_event, wait_event_type, query from pg_stat_activity where query like 'COPY%';
  pid  | state | wait_event | wait_event_type |                          query                           
-------+-------+------------+-----------------+----------------------------------------------------------
 25020 | idle  | ClientRead | Client          | COPY nycpluto_all TO stdout WITH(FORMAT csv,HEADER true)
 25028 | idle  | ClientRead | Client          | COPY nycpluto_all TO stdout WITH(FORMAT csv,HEADER true)
 25037 | idle  | ClientRead | Client          | COPY nycpluto_all TO stdout WITH(FORMAT csv,HEADER true)
 25045 | idle  | ClientRead | Client          | COPY nycpluto_all TO stdout WITH(FORMAT csv,HEADER true)
 25056 | idle  | ClientRead | Client          | COPY nycpluto_all TO stdout WITH(FORMAT csv,HEADER true)
(5 rows)
```

keep in mind that normally those idle connection/backends can be reused (if node pg knows they are available).

I'd rather write tests for this, but to be honest I don't know how to write sane ones for such scenarios.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@ test/test.log
 test/acceptance/oauth/venv/*
 coverage/
 npm-debug.log
+log/*.log
+yarn.lock
--- a/app/services/stream_copy.js
+++ b/app/services/stream_copy.js
@@ -34,8 +34,11 @@ module.exports = {
                         const runningClient = client;
                         const cancelingClient = new Client(runningClient.connectionParameters);
                         cancelingClient.cancel(runningClient, pgstream);
+
+                        const err = new Error('Connection closed by client');
                         pgstream.unpipe(res);
-                        done(new Error('Connection closed by client'));
+                        // see https://node-postgres.com/api/pool#release-err-error-
+                        done(err);
                         return cb(err);
                     }
                 })
@@ -45,7 +48,7 @@ module.exports = {
                 .on('error', err => {
                     if (!connectionClosedByClient) {
                         pgstream.unpipe(res);
-                        done(new Error('Connection closed by client'));
+                        done();
                         return cb(err);
                     }
                 })
--- a/app/services/stream_copy.js
+++ b/app/services/stream_copy.js
@@ -16,6 +16,9 @@ module.exports = {
             }
 
             let responseEnded = false;
+            let connectionClosedByClient = false;
+            const copyToStream = copyTo(sql);
+            const pgstream = client.query(copyToStream);
 
             res
                 .on('error', err => {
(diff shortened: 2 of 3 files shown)


rafatower:
A similar behavior happens with COPY FROM :cry:

oleurud:
After a big research days, and closing the week, I want to add more context to this problem.

First of all, we are trying to stop the copy query when a client closes the connection. Usually, we don't spend a lot of time on it, but this is a special query (that can use many resources). If we forget it, we won't have the current problem.

Following the Postgres documentation https://www.postgresql.org/docs/current/static/protocol-flow.html#PROTOCOL-COPY, in the COPY FROM case, we have solved the case sending a `CopyFail` message, it is not possible in the COPY TO case: 

```
The frontend cannot abort the transfer (except by closing the connection or issuing a Cancel request), but it can discard unwanted CopyData and CopyDone messages.
```

We have tried to cancel the request (the current state of this PR), but we have seen a random Postgres bad behavior. Let me explain it:
- THE USER closes the connection
- POSTGRES send copyOutResponse (0x48)
- WE send a cancel query request
- POSGTGRES send an ErrorResponse (0x45)
- WE release the connection to the db pool
- POSTGRES send a readyForQuery (0x5a)

But sometimes, we aren't receiving the `ErrorResponse`, so we are not releasing the connection to the db pool (so the user is losing a connection).

The other option, `closing the connection`, not seems an easy option in our stack.

oleurud:
I think that we are spending a lot of effort in something that it might not be a real problem. For sure there will be users stopping the COPY TO, but there could be a few queries. Also, with the current statement timeout limits, the queries will use 5, 15 or 25 seconds.

I would prefer to add more metrics, allowing us to measure the real behavior of the users, and then, with data, take care of the problems. 

Anyway, the current PR has some needed code (the `done()` requests for example) that we need not forget.

What do you think @rafatower?

rafatower:
After the sync meeting with rochoa and sancha the conclusion is that we're gonna focus on the COPY FROM.

All the problems with timeouts and disconnections lie in the COPY TO. I'm focusing right now on getting the tests stable. But I'll time box my work on the COPY TO.

I've run and checked manually the timeout of the `copyto` and the problem is around this changeset 994e8a702bd6fda13e42453821bdf621fe4a4e1e :
```
$ git bisect bad
994e8a702bd6fda13e42453821bdf621fe4a4e1e is the first bad commit
commit 994e8a702bd6fda13e42453821bdf621fe4a4e1e
Author: Rafa de la Torre <rtorre@carto.com>
Date:   Thu May 31 18:59:17 2018 +0200

    Add callbacks to pg.connect
    
    And call them to return connections to the pool.

:040000 040000 95416593ee305cf2be0a3c5d9431f8d363f9f96d 13b6e62ffe38fa029287006644d43ff657fc5cd8 M      app
```

so, most likely my fault, messing up with error management, all those `done` and the pg pool.

rafatower:
From my point of view, this is ready to merge in `pgcopy-stream` and deploy to staging. We know it's not perfect but it has the "right" tradeoffs.

Thanks for your help @oleurud

oleurud:
:muscle:

PR #499, opened 2018-06-06
Upgrade vulnerable modules

enekid:
There are currently some high and medium severity vulnerabilities in the `express` and `request` versions that we are using:

https://snyk.io/test/github/CartoDB/CartoDB-SQL-API/9912dbebad7aed6c3f346538b653ecd049ecf3b4

With this PR we get this result:

https://snyk.io/test/github/CartoDB/CartoDB-SQL-API/9160242858b04c74f72a768215507c0a9f883a26

There are too some medium vulnerabilities in the `node-uuid` version that `oauth-client` uses, but there is currently no `oauth-client` version that uses the fixed `node-uuid`, so I've created a PR:

https://github.com/unscene/node-oauth/pull/15


Algunenano:
I think we could go for express 4.16.3 and request 2.87.0. We are already using them in Windshaft-cartodb

Issue #498, opened 2018-06-05
Return exact Pg field type information in JSON return


pramsey:
**Motivation**: Right now clients can only infer generic type information (see below) about return fields, and they sometimes need/want more exact type information than we provide. They used to just look at the system tables to figure this stuff out, but now with read-only keys that restrict system table access, they really need to get this information directly from the result set. 

**End State**: Clients should be able to get a fast and precise table column/type definition just by running `SELECT * FROM mytable LIMIT 0`.

Right now we return loose field information:
```
   "fields" : {
      "the_geom" : {
         "type" : "geometry"
      },
      "created_at" : {
         "type" : "date"
      },
      "the_geom_webmercator" : {
         "type" : "geometry"
      },
      "id" : {
         "type" : "number"
      },
      "updated_at" : {
         "type" : "date"
      },
      "cartodb_id" : {
         "type" : "number"
      }
   }
```
But we could return the exact type information, which OGR uses to form the field definitions when creating OGRLayer objects in the CARTO driver. The [OGR types](http://www.gdal.org/ogr__core_8h.html#a787194bea637faf12d61643124a7c9fc) distinguish between real/integer, date and datetime. They also distinguish in geometries between the [geometry types](http://www.gdal.org/ogr__core_8h.html#a800236a0d460ef66e687b7b65610f12a) and the dimensionality of geometries (2d/3d) and the SRS of geometries.

We can extract the type/dim/srid information from the typmod number returned in libpq via `PQfmod(const PGresult *res, int field_num)`, as it all lives in [various bitfields](https://github.com/postgis/postgis/blob/1277c750cb41807eefee4438a711746042d0cffe/liblwgeom/liblwgeom.h.in#L164-L172) of the integer. 

The exact postgresql types can be inferred from the Oid, there's probably a lookup for this already behind the `fields` output. We just need to be a little more exact. Either go for a `fields_ogr` output list, or add a `type_ogr` entry to the existing `fields` or more generically just add a `pgtype` field.

```
   "fields" : {
      "the_geom" : {
         "type" : "geometry",
         "wkbtype": "Point",
         "dims": 2,
         "srid": 4326
      },
      "created_at" : {
         "type" : "date",
         "pgtype": "timestamptz"
      },
      "the_geom_webmercator" : {
         "type" : "geometry",
         "wkbtype": "Point",
         "dims": 2,
         "srid": 3857
      },
      "id" : {
         "type" : "number",
         "pgtype": "int4"
[...]

PR #500, opened 2018-06-06
WIP: Return 'pgtype' and other field info in JSON return?

pramsey:
To address #498, a PoC that returns extra, more exact field type information, hopefully to make things easier for OGR in reading tables without needing full administrative api key access.

### DIFF
--- a/app/models/formats/pg/json.js
+++ b/app/models/formats/pg/json.js
@@ -44,8 +44,8 @@ JsonFormat.prototype.formatResultFields = function(flds) {
     var rv = { type: tname, pgtype: cname };
 
     if ( cname.match(/geometry|geography/) ) {
-        var typmodInfo = this.client.typeModInfo(f.dataTypeModifier);
-        _.extend(rv, typmodInfo);
+      var typmodInfo = this.client.typeModInfo(f.dataTypeModifier);
+      _.extend(rv, typmodInfo);
     }
     nfields[f.name] = rv
   }
--- a/app/models/formats/pg/json.js
+++ b/app/models/formats/pg/json.js
@@ -41,7 +41,13 @@ JsonFormat.prototype.formatResultFields = function(flds) {
         tname += '[]';
       }
     }
-    nfields[f.name] = { type: tname };
+    var rv = { type: tname, pgtype: cname };
+
+    if ( cname.match(/geometry|geography/) ) {
+        var typmodInfo = this.client.typeModInfo(f.dataTypeModifier);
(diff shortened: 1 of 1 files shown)


Algunenano:
I've merged https://github.com/CartoDB/CartoDB-SQL-API/pull/493 which should make `typeModInfo` available. You'll need to rebase on top of the current master.

rafatower:
I rebased it on top of current `origin/master`. I guess this is now ready for a quick review.

rafatower:
@dgaubert ping

test/acceptance/stream-responses.js
```diff
@@ -39,8 +39,8 @@ describe('stream-responses', function() {
                     var parsedBody = JSON.parse(res.body);
```
enekid: The `describe` and `it` descriptions are a copy/paste from the next test suite.

`
describe('format-json', function() {
`
`
it('should close on error and error message must be part of the response'
`

app/models/formats/pg/json.js
```diff
@@ -41,7 +41,13 @@ JsonFormat.prototype.formatResultFields = function(flds) {
         tname += '[]';
       }
     }
-    nfields[f.name] = { type: tname };
+    var rv = { type: tname, pgtype: cname };
```
oleurud: I feel more comfortable with more verbose naming

pramsey: Like "pgsql_type" or "postgres_type"? (_ or -)?

enekid: what is `rv`?
Does github now charge for the use of vowels?

pramsey: Oh, you mean `rv` which is a musclememory name I always use for a throwaway "return value". Easy enough to change...

oleurud: postgres_type sounds nice :)

app/models/formats/pg/json.js
```diff
@@ -41,7 +41,13 @@ JsonFormat.prototype.formatResultFields = function(flds) {
...
+
+    if ( cname.match(/geometry|geography/) ) {
+      var typmodInfo = this.client.typeModInfo(f.dataTypeModifier);
+      _.extend(rv, typmodInfo);
```
dgaubert: What about `Object.assign(rv, typmodInfo)`?. I'd like to reduce the use of userland modules whenever is possible.

pramsey:
ok @dgaubert are we closer now?

PR #503, opened 2018-06-08
Pgcopy stream refactor2

oleurud:
Solves: #506 

Also, I have removed the old metrics in the headers


oleurud:
It is a first approach to the changes requested in #480 

I need to test it a lot before continue. But I suppose that it possible to review if it fits with the changes requested. I am not sure if it is better than the original. 

@dgaubert @enekid @rafatower

app/controllers/copy_controller.js
```diff
@@ -125,12 +213,12 @@ function validateCopyQuery () {
         const sql = req.query.q;
 
         if (!sql) {
-            next(new Error("SQL is missing"));
+            return next(new Error("SQL is missing"));
         }
 
         // Only accept SQL that starts with 'COPY'
         if (!sql.toUpperCase().startsWith("COPY ")) {
```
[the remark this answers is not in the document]
oleurud: I will remove the comment :joy:

rafatower:
As per our chat, closing this for the moment (but keeping the branch and all). Another issue/PR will be created instead. Feel free to reopen if needed.

oleurud:
I am going to reopen this PR, because I am going to follow with the work made here, changing the base branch

oleurud:
I have developed some different solutions:
- the current one: the event emitter
- double callback: it was into the commit https://github.com/CartoDB/CartoDB-SQL-API/pull/503/commits/8a15b2f8d10d096bedd43ee119ce8258dabebae9 The main difference is: Stream Copy, instead of emitting an event at the end with the results, receives a second callback and call it with the results.

Anyway, I prefer the current solution in master branch.

app/controllers/copy_controller.js
```diff
@@ -135,11 +132,24 @@ function handleCopyFrom (logger) {
...
         const isGzip = req.get('content-encoding') === 'gzip';
 
+        const streamCopy = new StreamCopy(sql, userDbParams);
         const metrics = new StreamCopyMetrics(logger, 'copyfrom', sql, user, isGzip);
 
+        streamCopy.on('copy-from-end', rows => {
+            metrics.end(rows);
+
+            const { time } = metrics;
+            if (!time || !rows) {
```
rafatower: Why would it have rows but time = zero?

oleurud: I expect if time == 0 / false / null / undefined something goes wrong. But yes, it is not a good implementation, because of the user could have inserted rows and we will return "No rows copied"

rafatower:
are you sure you wanted to merge this?

oleurud:
It was merged automatically when I merged https://github.com/CartoDB/CartoDB-SQL-API/pull/508 (#508 was created from this)

I came here to remove it and I have found it as "merged"

rafatower:
ha! wonders of github. Makes sense if it was the departing point of the final refactor. It's ok, let's leave it as it is. Thanks!

PR #509, opened 2018-06-14
Adding things from discarded refactors of COPY and some improvements

oleurud:
- Adding things from discarded refactors
- tests more deterministics
- `success` and `cdb-user` in logs
- logs rotate support

cc/ @dgaubert  @rafatower

### DIFF
--- a/app/controllers/copy_controller.js
+++ b/app/controllers/copy_controller.js
@@ -102,9 +102,9 @@ function handleCopyFrom (logger) {
                 }
 
                 // TODO: remove when data-ingestion log works
-                if (req.profiler) {	
-                    req.profiler.add({copyFrom: { type, format, gzip: isGzip, size, rows, time }});	
-                    res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());    	
+                if (req.profiler) {
+                    req.profiler.add({copyFrom: { type, format, gzip: isGzip, size, rows, time }});
+                    res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());
                 }
 
                 res.send({
--- a/app/controllers/copy_controller.js
+++ b/app/controllers/copy_controller.js
@@ -93,24 +93,16 @@ function handleCopyFrom (logger) {
             res.locals.user,
             req.get('content-encoding') === 'gzip', 
             logger,
-            function(err, metrics) {  // TODO: remove when data-ingestion log works: {time, rows}
+            function(err, metrics) { 
                 if (err) {
                     return next(err);
                 } 
 
-                // TODO: remove when data-ingestion log works
-                const { time, rows, type, format, gzip, size } = metrics; 
-
-                if (!time || !rows) {
+                const { time, rows } = metrics;
+                if (!rows) {
                     return next(new Error("No rows copied"));
                 }
 
-                // TODO: remove when data-ingestion log works
-                if (req.profiler) {
-                    req.profiler.add({copyFrom: { type, format, gzip, size, rows, time }});
-                    res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());    
-                }
-                
                 res.send({
                     time,
                     total_rows: rows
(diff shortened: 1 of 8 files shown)


test/acceptance/copy-endpoints.js
```diff
@@ -19,377 +20,380 @@ const server = require('../../app/server')(statsClient);
...
+                    url: `http://${address}:${port}/api/v1/sql/copyfrom?${query}`,
+                    headers: { host: 'vizzuality.cartodb.com' },
+                    method: 'POST',
+                    data: fs.createReadStream(__dirname + '/../support/csv/copy_test_table.csv')
+                };
+
+                const req = request(options);
+
+                setTimeout(() => {
+                    req.abort();
```
rafatower: hey, I got an idea about this one: IIRC, the client receives a 100 Continue before the stream of data, if you can capture that one and trigger upon reception, you can make a robust test. Don't know the exact implementation details, though.

oleurud: It seems not possible because of the `100 Continue` is sent before the client sends the request body. 

From https://www.w3.org/Protocols/rfc2616/rfc2616-sec8.html:
`
The purpose of the 100 (Continue) status (see section 10.1.1) is to allow a client that is sending a request message with a request body to determine if the origin server is willing to accept the request (based on the request headers) before the client sends the request body. In some cases, it might either be inappropriate or highly inefficient for the client to send the body if the server will reject the message without looking at the body.
`

app/controllers/copy_controller.js
```diff
@@ -140,6 +136,7 @@ function validateCopyQuery () {
 function errorHandler () {
     return function errorHandlerMiddleware (err, req, res, next) {
         if (res.headersSent) {
+            console.error("EXCEPTION REPORT: " + err.stack);
```
rafatower: Is this kept on purpose? (personally, I'm totally for it. Good to make sure that if you go to the log files you'll see full traces).

app/services/stream_copy_metrics.js
```diff
@@ -72,8 +73,11 @@ module.exports = class StreamCopyMetrics {
 
         if (errorMessage) {
             logData.error = errorMessage;
+            this.success = false;
```
rafatower: Do we log the `success: [true|false]`?

app/controllers/copy_controller.js
```diff
@@ -93,24 +90,23 @@ function handleCopyFrom (logger) {
...
                     return next(new Error("No rows copied"));
                 }
 
                 // TODO: remove when data-ingestion log works
-                if (req.profiler) {
-                    req.profiler.add({copyFrom: { type, format, gzip, size, rows, time }});
-                    res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());    
+                if (req.profiler) {	
+                    req.profiler.add({copyFrom: { type, format, gzip: isGzip, size, rows, time }});	
+                    res.header('X-SQLAPI-Profiler', req.profiler.toJSONString());
```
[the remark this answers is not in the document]
oleurud: Sure!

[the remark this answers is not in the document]
oleurud: Only a comment: since I am using vscode, I have been adding trailing whitespaces. But vscode has an option to remove them auto:
```
files.trimTrailingWhitespace": true
```

app/controllers/copy_controller.js
oleurud: In the usual way, when we have an error, the error middleware handles it. One of the task made by error handler is to log the error https://github.com/CartoDB/CartoDB-SQL-API/blob/master/app/middlewares/error.js#L14

As it is an special case and we can't use the error middleware, I have added it in the same way.

app/services/stream_copy_metrics.js
oleurud: We have `this.success = true` by default
https://github.com/CartoDB/CartoDB-SQL-API/pull/509/files/30402f2e890e2f38dd2900383962312d89a48075#diff-49b3b3377623e1ffd408f096a77c1610R19

And we always log the result with `logData.success = this.success;`

PR #510, opened 2018-06-20
Copy tests more deterministics

oleurud:
In order to get it, the idea (by @dgaubert) was:
- copy to: use a query with an insert and returning the inserted rows. If the copy to is canceled, the rows should not be in the database
- in copy form: the same idea: if the copy form query is canceled the rows should not be in the database

Other improvements:
- Ensuring the copy tests are independent

### DIFF
--- a/test/acceptance/copy-abort.js
+++ b/test/acceptance/copy-abort.js
@@ -43,7 +43,7 @@ function countInsertedRows (host, port, callback) {
     }, 1000);
 }
 
-describe.only('Cancel "copy to" commands', function () {
+describe('Cancel "copy to" commands', function () {
 
     beforeEach(function (done) {
         this.listener = server.listen(0, '127.0.0.1');
--- a/test/acceptance/copy-abort.js
+++ b/test/acceptance/copy-abort.js
@@ -1,0 +1,229 @@
+const querystring = require('querystring');
+const StatsClient = require('../../app/stats/client');
+const statsClient = StatsClient.getInstance(global.settings.statsd);
+const server = require('../../app/server')(statsClient);
+const request = require('request');
+const assert = require('assert');
+
+const copyQuery = `COPY (
+    INSERT INTO copy_to_test
+    SELECT updated_at
+    FROM generate_series('1984-06-14 01:00:00'::timestamp, '2018-06-14 01:00:00'::timestamp, '1 hour'::interval) updated_at
+    RETURNING updated_at
+) TO STDOUT`;
+
+const createTableQuery = `CREATE TABLE copy_to_test AS
+    (SELECT '2018-06-15 14:49:05.126415+00'::timestamp AS updated_at)`;
+
+const dropTableQuery = `DROP TABLE copy_to_test`;
+
+const countQuery = `SELECT count(1) as count FROM copy_to_test`;
+
+function countInsertedRows (host, port, callback) {
+    setTimeout(function () {
+        const count = querystring.stringify({ q: countQuery, api_key: 1234 });
+
+        const options = {
+            url: `http://${host}:${port}/api/v1/sql?${count}`,
+            headers: { host: 'vizzuality.cartodb.com' },
+            method: 'GET'
+        };
+
+        request(options, function (err, res, body) {
+            if (err) {
+                return callback(err);
+            }
+
+            assert.equal(res.statusCode, 200);
+
+            const result = JSON.parse(body)
+
+            callback(null, result);
+        });
+    }, 1000);
+}
+
+describe.only('Cancel "copy to" commands', function () {
+
+    beforeEach(function (done) {
+        this.listener = server.listen(0, '127.0.0.1');
+
+        this.listener.on('error', done);
+
+        this.listener.on('listening', () => {
+            const { address, port } = this.listener.address();
+
+            this.host = address;
+            this.port = port;
+
+            done();
+        });
+    });
+
+    beforeEach(function (done) {
+        const { host, port } = this;
+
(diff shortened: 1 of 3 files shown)


test/acceptance/copy-abort.js
```diff
@@ -0,0 +1,231 @@
...
+        request(options, function (err, res, body) {
+            if (err) {
+                return callback(err);
+            }
+
+            assert.equal(res.statusCode, 200);
+            const result = JSON.parse(body);
+            callback(null, result);
+        });
+    }, 1000);
```
rafatower: One second. Mind that this is a bit arbitrary, to give enough time for the `abort()` below.

test/acceptance/copy-abort.js
```diff
@@ -0,0 +1,231 @@
...
+
+        const createTableOptions = {
+            url: `http://${host}:${port}/api/v1/sql?${createTable}`,
+            headers: { host: 'vizzuality.cartodb.com' },
+            method: 'GET'
+        };
+
+        request(createTableOptions, function (err, res) {
+            if (err) {
+                return done(err);
```
rafatower: what does `done(err)` do in this `beforeEach` context?

rafatower: answering myself: https://mochajs.org/#asynchronous-hooks

test/acceptance/copy-abort.js
```diff
@@ -0,0 +1,231 @@
...
+
+        const dropTable = querystring.stringify({ q: dropTableQuery, api_key: 1234 });
+
+        const createTableOptions = {
+            url: `http://${host}:${port}/api/v1/sql?${dropTable}`,
+            headers: { host: 'vizzuality.cartodb.com' },
+            method: 'GET'
+        };
+
+        request(createTableOptions, function (err, res) {
```
rafatower: why do we need to create the table `afterEach`?

test/acceptance/copy-abort.js
```diff
@@ -0,0 +1,231 @@
...
+            }
+
+            assert.equal(res.statusCode, 200);
+
+            done();
+        });
+    });
+
+    afterEach(function (done) {
+        this.listener.close(done);
```
rafatower: can't there be a race condition between this `afterEach` and the one above, closing the listener too soon?

test/acceptance/copy-endpoints.js
```diff
@@ -82,55 +88,66 @@ describe('copy-endpoints', function() {
...
+                    assert.strictEqual(
+                        res.body,
+                        '11\tPaul\t10\n12\tPeter\t10\n13\tMatthew\t10\n14\t\\N\t10\n15\tJames\t10\n16\tJohn\t10\n'
+                    );
+
+                    assert.equal(res.headers['content-disposition'], 'attachment; filename=%2Ftmp%2Foutput.dmp');
+                    assert.equal(res.headers['content-type'], 'application/octet-stream');
+
+                    done();
+                });
```
rafatower: nice! :slightly_smiling_face:

test/acceptance/copy-endpoints.js
```diff
@@ -305,41 +322,54 @@ describe('copy-endpoints', function() {
...
             global.settings.db_pool_size = this.db_pool_size;
         });
-    
-        var assertCanReuseConnection = function (done) {
+
+        const assertCanReuseCanceledConnection = function (done) {
             assert.response(server, {
                 url: '/api/v1/sql?' + querystring.stringify({
-                    q: 'SELECT 1',
+                    q: 'SELECT count(*) FROM copy_endpoints_test',
```
rafatower: :+1: to the improved function name.

You know, the intent of `assertCanReuseConnection` was only to check that another query could reuse the same connection, thus it used the most simple query `SELECT 1`.

Now it does two things: it checks it can reuse the connection *and* it uses it for checking the row count. If this assertion ever fails, the person reviewing it can have a hard time figuring out what happened.

rafatower: It does not make much sense for the `COPY TO` test case. The `COPY TO` test case should have rows (and it has from the `generate_series()`)

test/acceptance/copy-abort.js
oleurud: :)

test/acceptance/copy-abort.js
oleurud: Ups, I should change the var name, because we are making a `drop table`

test/acceptance/copy-abort.js
oleurud: Yes, it is... and maybe too big. 

This is the reason why I set the title `more deterministic` instead of `deterministics`. We are assuming that 1 second is enough time to cancel the query.

test/acceptance/copy-abort.js
oleurud:
```
Hooks will run in the order they are defined
```
https://mochajs.org/#hooks

So, they will run in series and ordered.

test/acceptance/copy-endpoints.js
oleurud: :)

test/acceptance/copy-endpoints.js
oleurud: I agree. Do you feel more comfortable with 2 different functions?
- assertCanReuseCanceledConnection for copy from
- assertCanReuseConnection for copy to

test/acceptance/copy-abort.js
oleurud: We are creating it berofeEach and dropping it afterEach, as a way to make the tests independent.

PR #514, opened 2018-06-27
Improve sql timeout error message

### LINKED ISSUE #511: Improve message description for limit errors (a.k.a 429 Too many Request)
We need to improve this error message, to make sure that the user understands either his query timed out or he reached his `rate-limit`.
```
{
  error: [
    "You are over platform's limits. Please contact us to know more details"
  ],
  context: "limit",
  detail: "datasource"
}
```

enekid:
closes #511

When a DB timeout error arises caused by our `limits system`, the user should be aware of the actual error reason. Currently, the API responses with a cryptic 429 http error:

```json
{
    "error": [
        "You are over platform's limits. Please contact us to know more details."
    ],
    "context": "limit",
    "detail": "datasource"
}
```

This PR simply adds ` SQL query timeout expired error.` to the error message. This way, we don't break current users implementations (_hello frontenders!_):

```json
{
    "error": [
        "You are over platform's limits. Please contact us to know more details. SQL query timeout expired error."
    ],
    "context": "limit",
    "detail": "datasource"
}
```

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 ## 2.1.1
 Released 2018-mm-dd
 
+Announcements:
+  * Improve error message when the DB query is over the user's limits
 
 ## 2.1.0
 Released 2018-06-13
--- a/app/services/error_handler_factory.js
+++ b/app/services/error_handler_factory.js
@@ -18,7 +18,7 @@ function isTimeoutError(err) {
 
 function createTimeoutError() {
     return new ErrorHandler({
-        message: 'You are over platform\'s limits. Please contact us to know more details.' +
+        message: 'You are over platform\'s limits. Please contact us to know more details. ' +
                  'SQL query timeout expired error.',
         context: 'limit',
         detail: 'datasource',
--- a/app/services/error_handler_factory.js
+++ b/app/services/error_handler_factory.js
@@ -18,7 +18,8 @@ function isTimeoutError(err) {
 
 function createTimeoutError() {
     return new ErrorHandler({
-        message: 'You are over platform\'s limits. Please contact us to know more details',
+        message: `You are over platform\'s limits. Please contact us to know more details. 
+                  SQL query timeout expired error.`,
         context: 'limit',
         detail: 'datasource',
         http_status: 429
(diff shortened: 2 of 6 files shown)


andy-esch:
Or reversing the first sentence:
> SQL query timeout error: you are over the platform limits. Refactor your query before running again or contact CARTO support for more details.

enekid:
I've tried to not break any current integrations. Some of them rely on the message to identify the error (wrong on our part, we should provide an error code for that). For example, a lot of our own tests broke when changing the message.

I also like the @andy-esch message proposal.

@ivanmalagon Do you frontenders use this message in order to identify the error?

oleurud:
^^^ @rubenmoya

dgaubert:
> I've tried to not break any current integrations. Some of them rely on the message to identify the error (wrong on our part, we should provide an error code for that). For example, a lot of our own tests broke when changing the message.

Apart from out test, what do we break if we change the error's message?. Don't hesitate to change accordingly the message in our test.

rubenmoya:
AFAIK in Carto.js we're using [this RegEx](https://github.com/CartoDB/carto.js/blob/68b12d9acff88a65c872809b162d82c931a3e7e2/src/api/v4/error-handling/error-list/windshaft-errors.js#L22) to check the error type, so the new message should work too.

enekid:
> Apart from our test, what do we break if we change the error's message?. Don't hesitate to change accordingly the message in our test.

For example the `carto.js` integration 
I assume that a lot of users use these error messages as error codes via a regex.

I propose this message based on @andy-esch that honors carto.js regex:

```
You are over platform's limits: SQL query timeout error. Refactor your query before running again or contact CARTO support for more details.
```

I'm looking forward the global error refactoring 😄

oleurud:
Should we improve the rate limit message too?

dgaubert:
🙏 update NEWS.md before merging!

ivanmalagon:
With the proposed new error in the PR we won't have any problem. Our regex to categorize the error checks if the message starts with `You're over platform...`

If in the future the message doesn't start with that text, we should change our regex, but it's OK right now with the changed error text.

PR #516, opened 2018-07-05
Unnecessary read restriction

### LINKED ISSUE #513: Read restriction on `geometry_columns` is unnecessary
I see we are being pretty hard-core about preventing leakage from low-level access keys, by, for example disallowing access to `pg_*` tables that might ordinarily be visible to read users.

However, for `geometry_columns` this is unnecessary (I think) because the views will only allow users to see tables to which they have read permission. I demonstrated this with a test user on my own database.
```
CREATE USER test LOGIN;
GRANT SELECT ON foobar TO test;
```
Then connect as `test` and run `SELECT * FROM geometry_columns`. Only the `foobar` table shows up.

In general though, even this is not quite idea as we really want to allow read users to programatically list every *table* they have access to, not just every spatial table.

dgaubert:
Fixes #513

### DIFF
--- a/test/acceptance/export/involved-tables.js
+++ b/test/acceptance/export/involved-tables.js
@@ -1,0 +1,101 @@
+require('../../helper');
+
+const server = require('../../../app/server')();
+const assert = require('../../support/assert');
+const querystring = require('querystring');
+
+describe('Read restriction on "geometry_columns"', function() {
+    describe('OGR formats', function () {
+        it('GET /api/v1/sql downloads KML file from a private table', function (done){
+            const query = querystring.stringify({
+                api_key: 1234,
+                q: "SELECT * FROM private_table  LIMIT 1",
+                format: "kml",
+                filename: 'private_table'
+            });
+            assert.response(server, {
+                url: '/api/v1/sql?' + query,
+                headers: { host: 'vizzuality.cartodb.com' },
+                method: 'GET'
+            },{ }, function(err, res){
+                assert.equal(res.statusCode, 200, res.body);
+
+                const cd = res.headers['content-disposition'];
+
+                assert.ok(/filename=private_table.kml/gi.test(cd), cd);
+                assert.equal(res.headers['content-type'], 'application/kml; charset=utf-8');
+
+                done();
+            });
+        });
+
+        it('GET /api/v1/sql downloads a KML file from a public table', function (done){
+            const query = querystring.stringify({
+                q: "SELECT * FROM populated_places_simple_reduced LIMIT 1",
+                format: "kml",
+                filename: 'populated_places_simple_reduced'
+            });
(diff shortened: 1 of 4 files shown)


test/acceptance/export/involved-tables.js
```diff
@@ -0,0 +1,101 @@
...
+                const cd = res.headers['content-disposition'];
+
+                assert.ok(/filename=populated_places_simple_reduced.kml/gi.test(cd), cd);
+                assert.equal(res.headers['content-type'], 'application/kml; charset=utf-8');
+
+                done();
+            });
+        });
+
+        it('GET /api/v1/sql downloads a KML file from "geometry_columns" view', function (done){
```
dgaubert: Hey @pramsey!

I've started fixing #513, but I realized that I'm not sure which is the real use case that you want to be tackled. OGR driver only works when you want to download specific formats, such as: `kml`, `csv`, `shp`, etc... When you try to export a `kml` file of a private table w/o `api_key` it fails. If you send the request w/ your `api_key` it works. And if you want to export data from a public table you can request it with and without the `master api_key`. So far so good, no system tables are involved for public tables when you use the OGR driver.

When I try to send the folowing query w/o `api_key` (see this tests ^^): `SELECT * FROM geometry_columns LIMIT 1` it fails with `system tables are forbidden`. I understand that this is the use case that raises this issue, doesn't it?

If not, please help to me to understand better this issue. I've implemented a few tests with the examples that I just descibed.

Thanks in advance.

pramsey: For folks using OGR in a "not CARTO" context, the first thing OGR will do is ask CARTO, "hey, what tables are there?" (this is what happens when an OGRDataSource is instantiated). There's even a [super ugly hack](https://github.com/OSGeo/gdal/blob/0f195919394e29527e2f0d51eb3b5e778fb32795/gdal/ogr/ogrsf_frmts/carto/ogrcartodatasource.cpp#L231-L259) in OGR where it tries to sneak a peak at the system tables without letting on to CARTO that it is doing so. 

So, running `EXPORT CARTO_API_KEY=dfadfa ogrinfo CARTO:pramsey` should return to e a list  of tables I can read from, for example. Similarly, if I'm using an OGR driver to feed QGIS, when I pop open the layer selector I should see all the layers I have read access to.

Now, as it stands we're not going to re-open access to the pg_* tables, but the geometry_columns view is already a way to get a list of layers that are visible to the current user, so that's not a bad thing to open up.

Anyways, the use case is "given this key, what can I see?" and the apps that use it are things like ogrinfo and any other "select the thing you want" style application. Otherwise you're expecting the user to arrive with an _a priori_ list of things she has access to, which may not be the case.

dgaubert:
check

PR #517, opened 2018-07-05
Upgrading cartodb-redis

oleurud:
Updating cartodb-redis package:

fix bad behavior with an endpoint with several rate limits. More info: CartoDB/node-cartodb-redis#26
new major version removing deprecated functions CartoDB/node-cartodb-redis#28


oleurud:
It was already ready to use the new version (it is not using the deprecated functions)

package.json
```diff
@@ -23,12 +23,12 @@
     "bunyan": "1.8.1",
     "cartodb-psql": "0.12.0",
     "cartodb-query-tables": "0.2.0",
-    "cartodb-redis": "git://github.com/CartoDB/node-cartodb-redis.git#remove-auth-fallback",
+    "cartodb-redis": "^2.0.1",
```
enekid: Should we use:

`"cartodb-redis": "^2.0.1",`

Or:

`"cartodb-redis": "2.0.1",`

Does it matter if it's a Carto module or 3rd party?

🍈

oleurud: Yeep, I am going to change it

PR #519, opened 2018-07-18
Fix typo in example: s/query/q/

rafatower:
For the endpoint `api/v2/sql` the parameter to pass queries is `q`.
It is important to fix because it is in the SQL API front page. Thanks
@ibrahimmenem for spotting it.

### DIFF
--- a/docs/reference/swagger.yaml
+++ b/docs/reference/swagger.yaml
@@ -146,7 +146,7 @@ paths:
         - lang: Curl
           source: |
             curl -X POST -H "Content-Type: application/json" -d '{ \
-              "query": "SELECT count(*) FROM cities", \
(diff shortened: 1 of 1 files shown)


rafatower:
PS: There are other examples related to Batch API that use a `query` parameter. I assume that's correct, though slightly inconsistent:

https://github.com/CartoDB/CartoDB-SQL-API/blob/dce803d04eade109326f23bb6c37aaace0f2c288/app/controllers/job_controller.js#L125

enekid:
This PR fixes only the [SQL API reference section (swagger file)](https://carto.com/developers/sql-api/reference/#operation/postSQLStatement). The static [main page](https://carto.com/developers/sql-api/) should also be fixed

PR #520, opened 2018-07-18
Allow tests to be executed against pgbouncer

rafatower:
pgbouncer does not support the `--client-min-messages` option. Actually
it fails connections if used like that with this somewhat cryptic
message:

```
psql: ERROR:  Unsupported startup parameter: options
```

In order to be able to execute tests against pgbouncer port (which is
desirable IMO), we either need to remove that option (with little to no
impact) or change the lines above to choose the batch API port from the
config.

Mind that this affects just test setup.


rafatower:
Note I did not change the file `config/environments/test.js.example` because we need to add support for it in travis as well.

But I could change my `config/environments/test.js`, which resulted in this issue: https://github.com/CartoDB/CartoDB-SQL-API/issues/521

Issue #521, opened 2018-07-18
Timeouts not working for exports


rafatower:
At lest the tests do not pass when I run against `pgbouncer`.

### Steps to reproduce

1. Run `pgbouncer` listening on port `6432`.
1. Modify the `config/environments/test.js` to set `module.exports.db_port = '6432';`
1. Apply the patch in https://github.com/CartoDB/CartoDB-SQL-API/pull/520 because otherwise the tests die in the setup phase (explained there).
1. Execute the tests like this:
```
$ make test 2>&1 | tee /tmp/test.out
```

cause at some point the diffs display the bodies containing SHP and other binary formats, which screw my terminal output, so that way I could inspect the `/tmp/test.out` file.

### Expected result
All the tests pass.

### Actual result

Here's an excerpt of the results I got:

```
  415 passing (38s)
  13 pending
  9 failing

  1) timeout export database Geopackage export exceeding statement timeout responds 429 Over Limits:

      Uncaught AssertionError: Invalid response status code.
     Expected: 429
     Got: 200
     Body: SQLite format [...]

  2) timeout export database KML export exceeding statement timeout responds 429 Over Limits:

      Uncaught AssertionError: Invalid response status code.
     Expected: 429
     Got: 200
     Body: <?xml version="1.0" encoding="utf-8" ?> [...]

  3) timeout export database Shapefile export exceeding statement timeout responds 429 Over Limits:

      Uncaught AssertionError: Invalid response status code.
     Expected: 429
     Got: 200
     Body: [...]
```

(all the failures are related to exports)

rafatower:
(assigning to @dgaubert , for you to prioritize and dispatch at your leisure)

dgaubert:
Hey!

That's because PGBoucer has its internal pool of connection against Postgres, before executing those tests we need to set a proper [timeout for that user/database](https://github.com/CartoDB/CartoDB-SQL-API/blob/master/test/support/test-client.js#L120-L150). it uses one connection of PGBouncer pool but rest of them keep the connection with the previous settings and the timeout will not be applied.

In fact, [we had to clean up connections](https://github.com/CartoDB/CartoDB-SQL-API/blob/master/test/support/test-client.js#L134) from the internal `cartodb-psql` pool.

PR #522, opened 2018-07-19
521 test robustness pgbouncer

rafatower:
This PR makes it possible to execute tests with and without pgbouncer.

This fixes the issue https://github.com/CartoDB/CartoDB-SQL-API/issues/521

### DIFF
--- a/test/support/db_utils.js
+++ b/test/support/db_utils.js
@@ -27,11 +27,10 @@ module.exports.resetPgBouncerConnections = function (callback) {
     client.query('PAUSE', (err, res) => {
         if (err) {
             return callback(err);
-        } else {
-            client.query('RESUME', (err, res) => {
-                client.end();
-                return callback(err);
-            });
         }
+        client.query('RESUME', (err, res) => {
+            client.end();
+            return callback(err);
+        });
     });
 }
--- a/test/support/db_utils.js
+++ b/test/support/db_utils.js
@@ -1,0 +1,37 @@
+'use strict';
+
+const { Client } = require('pg');
+
+const dbConfig = {
+    db_user: process.env.PGUSER || 'postgres',
+    db_host: global.settings.db_host,
+    db_port: global.settings.db_port,
+    db_batch_port: global.settings.db_batch_port
+};
+
+module.exports.resetPgBouncerConnections = function (callback) {
+    // We assume there's no pgbouncer if db_port === db_batch_port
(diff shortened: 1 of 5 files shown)


rafatower:
I understand that for this to be really more automated and useful, I'd need to tweak the travis build and docker image to make use of `pgbouncer`, another dependency, and that it might not be desirable.

No big deal and no strong opinion, it is simply that it helped me to run existing tests, taking into account subtle differences between direct vs pgbounced connections.

test/support/db_utils.js
```diff
@@ -0,0 +1,37 @@
...
+    }
+
+    const client = new Client({
+        database: 'pgbouncer',
+        user: dbConfig.db_user,
+        host: dbConfig.db_host,
+        port: dbConfig.db_port
+    });
+
+    // We just chain a PAUSE followed by a RESUME
```
oleurud: Ei, what is the objective of pause - resume?

rafatower: https://pgbouncer.github.io/usage.html#admin-console

> PAUSE [db];
> PgBouncer tries to disconnect from all servers, first waiting for all queries to complete. The command will not return before all queries are finished. To be used at the time of database restart.

> RESUME [db];
> Resume work from previous PAUSE or SUSPEND command.

Wait for all the running queries to finish and let the new ones reconnect. This way you pick the changed session parameters such as `set statement_timeout`. Sames happens in production, BTW, but there the connections are refreshed usually more often.

oleurud: Ok, thanks! :)

test/support/db_utils.js
```diff
@@ -0,0 +1,37 @@
...
+    client.connect();
+    client.query('PAUSE', (err, res) => {
```
dgaubert: Please, follow Node.js conventions:

```js
client.query('PAUSE', (err, res) => {
  if (err) {
    return callback(err);
  }
  client.query('RESUME', (err, res) => ...);
}
```
If  you inmediatly return after calling the callback, there is no need to wrap following sentences into a `else` statement.

test/support/db_utils.js
dgaubert: What about `// We just chain a PAUSE followed by a RESUME to reset internal pool connections of PgBouncer`?

rafatower:
I covered your comments (that I always appreciate) and pleased jshint. Now time to merge. Thanks!

oleurud:
:+1:

PR #525, opened 2018-07-26
Copy configs tweak

rafatower:
This PR does a couple of things, mostly for housekeeping:
- it adds the missing copy configs to `config/environments/staging.js.example`
- it sets the default COPY_FROM_MAX_POST_SIZE to exactly 2 GB.

It was previously set to 1.99 GB (0.5 % error, a small lie) in order to be able to report a meaningful error to the client, instead of having nginx report a different more obscure error.

Now the nginx limitation is removed for that endpoint, so it gives us the flexibility of modifying this limit at app level (chances are in a future it will be overridable by user).


rafatower:
related to https://github.com/CartoDB/cartodb-platform/issues/4765

PR #526, opened 2018-07-31
Fix copy docs

rafatower:
This does two things:
- clarifies the need of table creation and adds a ready-to-use curl snippet
- replaces all http to https in the examples

The second one is pretty important, cause people tend to copy/paste examples.

@zingbot I'd appreciate a quick review to make sure this reads properly.

cc/ @dgaubert for awareness (just docs but urgent imho, so probably already merged by the time you read it).

### DIFF
--- a/doc/copy_queries.md
+++ b/doc/copy_queries.md
@@ -4,8 +4,8 @@ Copy queries allow you to use the [PostgreSQL copy command](https://www.postgres
 
 The support for copy is split across two API end points:
 
-* `http://{username}.carto.com/api/v2/sql/copyfrom` for uploading data to CARTO
-* `http://{username}.carto.com/api/v2/sql/copyto` for exporting data out of CARTO
+* `https://{username}.carto.com/api/v2/sql/copyfrom` for uploading data to CARTO
+* `https://{username}.carto.com/api/v2/sql/copyto` for exporting data out of CARTO
 
 ## Copy From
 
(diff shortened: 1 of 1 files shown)


rafatower:
Thanks a lot for the review (beyond the changes).

> Would maybe replace 'terse' in the Python section with 'simple', or 'easy'. 'Terse' can be used in a lot of ways and usually negatively refers to a conversation.

Fair enough, the original wording of that is from Paul and I trusted his English :smile: 

> 'A slightly more sophisticated script could read the headers from the CSV and compose the COPY command on the fly.' We might consider (not necessarily for release) adding the beginnings of this? It might increase more advanced usage of COPy. It also goes to a principle I'm big on, which is showing slightly more advanced methods to make the functions clearer. Intermediate users will want to use this and we should show them how to grow.

No, it can not. The CSV files do not contain type information so it cannot be done that way. That's what the Import Guessing does with varying degrees of success, as you know. With the COPY we decided to give control to the user vs guessing.

I thought we were clear about that, because that was a basic assumption :cry: 

> Break this sentence into two parts. "The only complexity in this example is at the end, where the result is streamed back block-by-block, to avoid pulling the entire download into memory before writing to file." 
 > "The only complexity in this example is at the end, where the result is streamed back block-by-block. This is to avoid pulling the entire download into memory before writing to file."

OK. I'm trying to hide that implementation detail in the Python SDK, BTW.

> Rewiting this with nuanced changes:
"There is a 5-hour timeout limit for the /copyfrom and /copyto endpoints. The idea behind that limit is that, in practice, COPY operations should not be limited by your regular query timeout. [...]

OK

> We should tell them what to do if they have 4-5gb to upload. That will come up a lot and we should make it easy to navigate. As in, what are the next steps to append more data to the files. Explaining this clearly will save us many hours in support time.

What cases do you expect? from whom? Wait, do not rush into answering. Let's discuss that in this other issue, please: https://github.com/CartoDB/cartodb-platform/issues/4544

rafatower:
I'm merging this now, I'll pick your comments for another docs polishing round in another PR.

PR #530, opened 2018-08-10
Handle gunzip/zlib errors

### LINKED ISSUE #528: copyfrom `Uncaught exception: Error: incorrect header check`
If you request a gzipped copyfrom sending a bad formatted gzip body (at least, we think the cause is a bad formatted gzip body), the API crashes. In the log you can see: 
```
Uncaught exception: Error: incorrect header check
    at Zlib._handle.onerror (zlib.js:370:17)
```

The error was reported by @rafatower who could add more info

rafatower:
This fixes #528

### DIFF
--- a/app/controllers/copy_controller.js
+++ b/app/controllers/copy_controller.js
@@ -115,16 +115,16 @@ function handleCopyFrom (logger) {
 
             req
                 .on('data', data => isGzip ? metrics.addGzipSize(data.length) : undefined)
-                .on('error', err => {
-                    metrics.end(null, err);
-                    pgstream.emit('error', err);
-                })
                 .on('close', () => {
                     const err = new Error('Connection closed by client');
                     pgstream.emit('cancelQuery', err);
                     pgstream.emit('error', err);
                 })
                 .pipe(isGzip ? zlib.createGunzip() : new PassThrough())
+                .on('error', err => {
+                    metrics.end(null, err);
+                    pgstream.emit('error', err);
+                })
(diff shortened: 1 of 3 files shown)


app/controllers/copy_controller.js
```diff
@@ -115,16 +115,16 @@ function handleCopyFrom (logger) {
 
             req
                 .on('data', data => isGzip ? metrics.addGzipSize(data.length) : undefined)
-                .on('error', err => {
```
oleurud: We should maintain it because if an error happens in the request we will have the same problem

oleurud: By the way, the new error handler is correct. Good catch

rafatower: OK. Then if the error handler can be specific to `Gunzip` I'd decorate the error with something like: `"Error while gunzipping: {error}"`. E.g: `"Error while gunzipping: incorrect header check"`. That would make it easier to understand for client developers (it got me crazy for a while, because I did not get any error on the client to start with).

app/controllers/copy_controller.js
```diff
@@ -125,6 +125,11 @@ function handleCopyFrom (logger) {
                     pgstream.emit('error', err);
                 })
                 .pipe(isGzip ? zlib.createGunzip() : new PassThrough())
+                .on('error', err => {
+                    err.message = `Error while gunzipping: ${err.message}`;
```
rafatower: nice!

PR #531, opened 2018-08-14
create carto-package.json file

oleurud:
One of CartoDB/onpremises#575

### DIFF
--- a/carto-package.json
+++ b/carto-package.json
@@ -2,15 +2,14 @@
     "name": "carto_sql_api",
     "current_version": {
         "requires": {
-            "node": ">=6.9.2",
+            "node": ">=6.9.2 <10.0",
             "npm": ">=3.10.9"
         },
         "works_with": {
             "redis": ">=3.0.0",
-            "postgresql": "9.5.2",
-            "postgis": "2.2.2.0",
-            "mapnik": "",
-            "gdal":""
+            "postgresql": ">=9.5",
+            "postgis": ">=2.2",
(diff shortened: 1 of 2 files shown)


carto-package.json
```diff
@@ -0,0 +1,16 @@
+{
+    "name": "carto_sql_api",
+    "current_version": {
+        "requires": {
+            "node": ">=6.9.2",
+            "npm": ">=3.10.9"
+        },
+        "works_with": {
+            "redis": ">=3.0.0",
+            "postgresql": "9.5.2",
```
rafatower: Whatever we decide for postgres, postgis, mapnik and gdal in https://github.com/CartoDB/Windshaft-cartodb/pull/1019 apply here, if I'm not mistaken.

carto-package.json
```diff
@@ -0,0 +1,16 @@
...
+            "postgis": "2.2.2.0",
+            "mapnik": "",
+            "gdal":""
```
rafatower: Ok, this is used for export. If we picked the minimum required a 1.10 should be enough, but 2.2 or newer the recommendation (latest stable is 2.3.1, but not tested in production).
```
"gdal": ">=1.10"
```

oleurud:
We have `gdal` as dependency is `Windshaft-carto` and also, I thought it was a dependency here. But after checking it, I don't see it. 

Nevertheless, we install `gdal-bin` in Travis configuration. So, I am going to remove it to test if it is needed or not

Issue #515, opened 2018-07-02
COPY TO: Log when the client request compression


rafatower:
We have some traces from users requesting `COPY TO` operations, with compression in the logs.

We know for a fact that they requested compression, because there's a difference in reported sizes between SQL API traces and nginx traces, showing a compression ratio > 50%.

If the header `Accept-Encoding: gzip` is present, the field of the trace `data-ingestion.gzip` should be set to `true`. Wondering if that header is passed from nginx to node or if it is lost in the way.

cc/ @dppascual @oleurud

rafatower:
@CartoDB/rt-managers I think this is a good candidate for RT

oleurud:
PR https://github.com/CartoDB/CartoDB-SQL-API/pull/532

oleurud:
Adding more info about it. 

I decided to set the gzip value as `null` because of node doesn't know if the compression happens or not. 

Nginx is the responsible about compress the response and although the client sends the `Accept-Encoding: gzip`, nginx may choose not to compress the body https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding

Knowing that, for me, it has sense to save in the log as `true` or `false` because it gives us an important information.

PR #532, opened 2018-08-14
copyto adding gzip to log

oleurud:
Solves: #515


rafatower:
Let me know whenever you want to test it in a staging. I can help you with the python sdk regression tests. (we have kibana metrics in staging as well, don't we?)

oleurud:
I have just deployed it to staging cloud. And yes, we should have Kibana in staging

rafatower:
For the record: the quick check showed that regression tests pass in staging and that the metrics file is generated as intended (with the gzip field set to true when compression is requested)

Thanks!

PR #533, opened 2018-08-17
Update carto-package.json

lekum:
Versions fixed

### DIFF
--- a/carto-package.json
+++ b/carto-package.json
@@ -2,15 +2,15 @@
     "name": "carto_sql_api",
     "current_version": {
         "requires": {
-            "node": ">=6.9.2 <10.0",
+            "node": ">=6.9.2 <10.0.0",
             "npm": ">=3.10.9",
             "gdal":">=1.11.0"
         },
         "works_with": {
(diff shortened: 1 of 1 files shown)


carto-package.json
```diff
@@ -2,15 +2,15 @@
...
             "npm": ">=3.10.9",
             "gdal":">=1.11.0"
         },
         "works_with": {
             "redis": ">=3.0.0",
-            "postgresql": ">=9.5",
-            "postgis": ">=2.2",
-            "carto_postgresql_ext": ">=0.19"
+            "postgresql": ">=9.5.0",
+            "postgis": ">=2.2.0.0",
```
oleurud: do you support 4 digists?

lekum: Yes, we support them using a "hack" (changing `a.b.c.d` to `a.b.c-d`) but only for components that we cannot easily adapt the tagging (as `postgis`)

lekum:
May I merge it then?

oleurud:
Yes, you have my blessings!

PR #537, opened 2018-09-17
Fix paused stream (DB socket) after copy to

rafatower:
This PR does 2 things:
- It tweaks an existing test to make the issue reproducible (see the failures in the first commit)
- It fixes the underlying problem with a workaround.


rafatower:
Related to: https://github.com/brianc/node-pg-copy-streams/issues/24

PR #547, opened 2018-12-11
 Properly mock Date.now() for tests

rafatower:
Under node 10, some tests fail _sometimes_ because of the way `Date.now()` is mocked: https://github.com/CartoDB/CartoDB-SQL-API/blob/b9f647f15d3c23c8859fe1ad3d38e50193d965ab/test/acceptance/last-modified-header.js#L66-L70 (though they work fine under node 6).

This can be checked on current master by running this:

```
$ for i in $(seq 1 100); do echo $i; test/run_tests.sh --nocreate  test/acceptance/last-modified-header.js || break; done
...
  5 passing (181ms)
  1 failing

  1) last modified header should use Date.now() for functions or results with no table associated:

      Uncaught AssertionError [ERR_ASSERTION]: 'Tue, 11 Dec 2018 10:58:05 GMT' == 'Tue, 11 Dec 2018 10:58:04 GMT'
      + expected - actual

      -Tue, 11 Dec 2018 10:58:05 GMT
      +Tue, 11 Dec 2018 10:58:04 GMT
      
      at test/acceptance/last-modified-header.js:113:24
      at Server.<anonymous> (test/support/assert.js:108:24)
      at emitCloseNT (net.js:1618:8)
      at process._tickCallback (internal/process/next_tick.js:63:19)
```

(just by chance some request may lay on the second after `fixedDateNow` is fixed).

This fixes the issue by using [MockDate](https://github.com/boblauer/MockDate), instead of elaborating our own mock (quite simple and understandable code anyway), making the tests more robust.

I'm using node v10.14.1

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -1649,6 +1649,12 @@
         }
       }
     },
+    "mockdate": {
+      "version": "2.0.2",
+      "resolved": "https://registry.npmjs.org/mockdate/-/mockdate-2.0.2.tgz",
+      "integrity": "sha1-WuDA6vj+I+AJzQH5iJtCxPY0rxI=",
+      "dev": true
+    },
     "moment": {
       "version": "2.22.2",
       "resolved": "https://registry.npmjs.org/moment/-/moment-2.22.2.tgz",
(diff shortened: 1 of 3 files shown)


dgaubert:
It isn't related just to Node.js 10, this error in tests comes from time to time also in Node.js 6 or even in Node.js 0.10. Anyway, nice to see it finally fixed. :)

PR #569, opened 2019-02-06
batch needs master key in docs

oleurud:
Comes from https://github.com/CartoDB/support/issues/1932

Right now we are enforcing the use of Master API Key for Batch API https://github.com/CartoDB/CartoDB-SQL-API/blob/master/app/controllers/job_controller.js#L61 and we should reflect it in docs


alrocar:
Uppss, not the first repo where I've found duplicated doc folders 🤔 Should we just maintain one version (the public one) and get rid of the other?

dgaubert:
Yes, we should. But we cannot just merge and get rid of the `developer-center` branch as we use that branch to deploy the developer center documentation. So it's a bit more complicated than that.

oleurud:
Yep, I forgot to add it: https://github.com/CartoDB/CartoDB-SQL-API/pull/570

Issue #573, opened 2019-02-26
Improve SQL-API metrics


ethervoid:
Include the POST requests in our metrics. The main objective of this change is to have DDL commands included for the new Ghost tables feature

oleurud:
We will need to add a new object in the logs. The current logs written has the following format :
```
[:date] :remote-addr :method :req[Host]:url :status :response-time ms -> :res[Content-Type] (:res[X-SQLAPI-Profiler]) (:res[X-SQLAPI-Errors])
```

My proposal is to add a new JSON object at the end in ther same way as `X-SQLAPI-Errors`:
```
[:date] :remote-addr :method :req[Host]:url :status :response-time ms -> :res[Content-Type] (:res[X-SQLAPI-Profiler]) (:res[X-SQLAPI-Errors]) (:res[X-SQLAPI-Log])
```

The new object will be generic to allow us to add everything in this JSON object in the future. It is not an objective of the current issue to define a future log structure, but the new object should be generic and flexible enough.

```
{
    request: {
        q: "SELECT * FROM ..."
    }
}
```

How do you see it @dgaubert ?

This approach will require to make some task with the infrastructure team. I have not reviewed it yet, but if I don't remember badly, I think I will be able to prepare the PRs myself

oleurud:
We already have something that works: https://github.com/CartoDB/CartoDB-SQL-API/blob/master/app/server.js#L116 But we have the same problem as adding a new object, we need to make some task with the infrastructure team. 

In my opinion, I prefer to create a new object as I explained in the previous comment to be more generic and flexible for the future.

But I like a point from it: if for some reason we need to add the queries to the log, it is possible adding `:sql` to the log format parameter. So, I think it makes sense to add a config parameter to enable/disable the query logging

oleurud:
The infrastructure part: https://github.com/CartoDB/cartodb-platform/pull/5399

dgaubert:
Just one thing: 

As we are using `log4js` with the following format:

> [:date] :remote-addr :method :req[Host]:url :status :response-time ms -> :res[Content-Type] (:res[X-SQLAPI-Profiler]) (:res[X-SQLAPI-Errors])

That means we also have access to the request, its params and, its body, right?. No sure if it's usefule at all, just thinking out loud, but something like `:res[locals][query]` could work. Thus, we can avoid to save the query in a response header.

What do you think?

oleurud:
[...]

PR #574, opened 2019-02-26
Log queries

oleurud:
Solves https://github.com/CartoDB/CartoDB-SQL-API/issues/573

Tasks:
- Remove the old query system
- Added the parameter `logQueries` to enable or disable queries log
- Handle queries in middleware and save them in `res.locals.sql`
- Using the previous middleware in query, job, and copy controllers endpoints
- Generic log middleware (using headers system `X-SQLAPI-Log`). Now it is only used for queries log, but it could be the solution to change `string` log to `JSON` log
- Using log middleware only in query and job controllers endpoints. Copy endpoints use their log
- extracting `stringifyForLogs` function and refactor to avoid modifying the original


app/controllers/job_controller.js
```diff
@@ -120,11 +124,9 @@ function getJob (jobService) {
 
 function createJob (jobService) {
     return function createJobMiddleware (req, res, next) {
-        const params = Object.assign({}, req.query, req.body);
-
         var data = {
             user: res.locals.user,
-            query: params.query,
+            query: res.locals.sql,
```
ethervoid: Why use sql instead of query here?

ethervoid: Oh I see, that comes from the middleware

app/middlewares/log.js
```diff
@@ -0,0 +1,41 @@
+'use strict';
+
+const { stringifyForLogs } = require('../utils/logs');
+
+module.exports = function log() {
+    return function logMiddleware(req, res, next) {
+        const MAX_SQL_LENGTH = (global.settings.logQueries && global.settings.maxQueriesLogLength) || 1024;
```
ethervoid: Why did just limit to 1024?

ethervoid: Another question should we just `return next()` if `global.settings.logQueries` is false?

test/acceptance/log.js
```diff
@@ -0,0 +1,215 @@
...
+        global.settings.logQueries = logQueries;
+    });
+
+    describe('regular queries endpoint', function() {
+        ['GET', 'POST'].forEach(method => {
+            it(`${method} without query fails`, function(done) {
+                assert.response(server,
+                    {
+                        method,
+                        url: '/api/v1/sql?' + qs.stringify({
```
ethervoid: Silly question, what about the `api/v2/sql`? do we need to test it too?

app/middlewares/log.js
oleurud: The number here is "random". The important thing is the number is configurable via config parameters, so we can change easily without touching the code.

The reason for limit it, it is to avoid problems in the logs infrastructure. We know if the logs are too big, we will have problems processing them.

And about the last comment, my point was to make a generic and flexible for the future (not only for queries logging). But right now, how it is only used by the queries logging, so it could do it

test/acceptance/log.js
oleurud: Basically, from this code point of view, using `v1` or `v2` it is the same. Then, in our SaaS is not the same due to infrastructure-related decisions (cache)

ethervoid: :+1:

app/middlewares/log.js
ethervoid: :+1: to the first part

About the second answer is better to have a `disconnect` flag for this case and then in the future change it if needed

oleurud: I prefer to have it working. I mean, I understand your point, it will be enabled by default and it won't be a noise source (at least if we don't change our opinion about it)

ethervoid: So what is the purpose of that flag? then we should remove it and avoid unnecessary flags in our config

oleurud: The config parameter is for enable/disable the queries (to disable it if we are overloading the infrastructure for example), not to enable/disable all the logs from the generic log. I mean, I aggre with you if we only use it as queries logger ([we already had a queries logger](https://github.com/CartoDB/CartoDB-SQL-API/pull/574/commits/799b7675054473bd827f036b2633d9f20673ff52)), but it is (it will be) more than that, and dont want to use a config parameter for queries log to enable/disable all the logs.

Said that, since we have a way to control the queries size with `maxQueriesLogLength`, we could change the config parameter from `enable/disable log queries` to `enable/disable the full logger`. What do you think?

ethervoid:
Good work mate :)

PR #576, opened 2019-03-15
remove python from libraries support section in Dev Center

oriolbx:
Python SDK is already a separated library and maintained by CARTO


jsanz:
Removing myself from the review in favor of @dgaubert, the reasoning is that Python SDK is part of our official offering so it should removed from this section of the docs

PR #588, opened 2019-05-14
Improve copy[to|from] cancel mechanism

dgaubert:
Implement a cancel backend through query instead of `node-postgres` protocol.

### DIFF
--- a/package.json
+++ b/package.json
@@ -34,7 +34,7 @@
     "node-statsd": "~0.0.7",
     "node-uuid": "^1.4.7",
     "oauth-client": "0.3.0",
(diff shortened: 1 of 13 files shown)


package.json
```diff
@@ -34,7 +34,7 @@
     "node-statsd": "~0.0.7",
     "node-uuid": "^1.4.7",
     "oauth-client": "0.3.0",
-    "pg-copy-streams": "github:cartodb/node-pg-copy-streams#v1.2.0-carto.3",
+    "pg-copy-streams": "^2.2.0",
```
oleurud: :muscle:  :tada:

PR #589, opened 2019-05-24
Improve copy error handling

dgaubert:


app/services/stream_copy.js
```diff
@@ -31,6 +32,7 @@ module.exports = class StreamCopy {
     }
 
     getPGStream(action, callback) {
+        this.action = action;
```
oleurud: I think you are not using it

dgaubert: done!

PR #592, opened 2019-06-03
Retrieve exact Pg field type information in JSON format

dgaubert:
Implements #498

### DIFF
--- a/test/acceptance/stream-responses.js
+++ b/test/acceptance/stream-responses.js
@@ -41,8 +41,16 @@ describe('stream-responses', function() {
                     var parsedBody = JSON.parse(res.body);
                     assert.equal(parsedBody.rows.length, 2);
                     assert.deepEqual(parsedBody.fields, {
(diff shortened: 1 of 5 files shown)


test/acceptance/stream-responses.js
```diff
@@ -40,10 +40,6 @@ describe('stream-responses', function() {
                 function(err, res) {
                     var parsedBody = JSON.parse(res.body);
                     assert.equal(parsedBody.rows.length, 2);
-                    assert.deepEqual(parsedBody.fields, {
-                        the_geom: { type: "geometry" },
```
oleurud: Can not we add the new `pgtype`?

dgaubert: The test is about checking the error while streaming responses. IMHO, the assertion is checking `fields` which I think is out of the scope of the test.

oleurud: ok

PR #599, opened 2019-07-04
Set directive 'max-age' to a fallback value

dgaubert:
Set directive 'max-age' to a fallback value when there are affected tables where we can't know when were updated for the last time, e.g: non-cartodified tables or foreign tables without cartodb support


test/acceptance/cache-headers.js
```diff
@@ -0,0 +1,91 @@
...
+            },
+            method: 'GET'
+        },
+        {},
+        function (err, res) {
+            assert.equal(res.headers.vary, 'Authorization');
+            done();
+        });
+    });
+
```
Algunenano: Do we need to check also the `mayWrite` (expected 0) path?

dgaubert: yeah, I don't see any reason to not check that. On it...

dgaubert: Already checked [here](https://github.com/CartoDB/CartoDB-SQL-API/pull/599/files#diff-024b7a24b50a6b3e116b92670ef25605R45).

PR #600, opened 2019-07-05
 Remove old abort checker

dgaubert:
Regarding this piece of code:

```js
// client-abort.js
'use strict';

const http = require('http')
const port = 5555

const server = http.createServer((req, res) => {
  req.on('aborted', () => console.log(`Server: req aborted by client (req.aborted: ${req.aborted})`))
  setTimeout(() => res.end('chunk of data'), 100)
});

server.listen(port);

server.on('listening', () => {
  const req = http.get({ port });

  req.on('error', err => {
    console.error('Client: error', err)
    server.close()
  })

  setTimeout(() => req.abort(), 50)
})
```

```sh
$ node client-abort.js
Server: req aborted by client (req.aborted: true)
Client: error { Error: socket hang up
    at createHangUpError (_http_client.js:323:15)
    at Socket.socketCloseListener (_http_client.js:364:25)
    at Socket.emit (events.js:203:15)
    at TCP._handle.close (net.js:606:12) code: 'ECONNRESET' }
```

Instead, use a Node-ish mechanism to achieve the same goal

### DIFF
--- a/app/controllers/query_controller.js
+++ b/app/controllers/query_controller.js
@@ -22,6 +22,7 @@ const rateLimitsMiddleware = require('../middlewares/rate-limit');
 const { RATE_LIMIT_ENDPOINTS_GROUPS } = rateLimitsMiddleware;
 const handleQueryMiddleware = require('../middlewares/handle-query');
 const logMiddleware = require('../middlewares/log');
+const cancelOnClientAbort = require('../middlewares/cancel-on-client-abort');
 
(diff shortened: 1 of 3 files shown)


dgaubert:
I understand your point, it makes sense for people that are not familiar with express/node ecosystem. I tend to define meaningful names to avoid large explanations in comments. I'm not a big fan of comments but for documentation. I'd like to use JSDoc or similar, however I didn't find the time to give it a try.

PR #608, opened 2019-07-31
Set a hard limit on the size of the X-SQLAPI-Log header

Algunenano:
Prevents X-SQLAPI-Log having a high length when dealing with multiple queries:
- If there are N queries, each one will include at most `global.settings.maxQueriesLogLength / N` when printed.
- As N can be arbitrarily high and the header contains more characters than just the query, it sets a hard limit of 4096 characters for the header.

### DIFF
--- a/app/middlewares/log.js
+++ b/app/middlewares/log.js
@@ -3,6 +3,12 @@
 const { stringifyForLogs } = require('../utils/logs');
 
 const MAX_SQL_LENGTH = (global.settings.logQueries && global.settings.maxQueriesLogLength) || 1024;
+
+// This is used to set a hard limit in the header size
+// While Node accepts headers of up to 8192 character, but different libraries impose other limits
+// This might break the JSON structure of the log
+const HEADER_HARD_LIMIT = 4096;
+
 const TYPES = {
     QUERY: 'query',
     JOB: 'job'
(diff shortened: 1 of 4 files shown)


Algunenano:
@oleurud @dgaubert Can you have a look at this, please?

app/middlewares/log.js
```diff
@@ -16,7 +22,7 @@ module.exports = function log(sqlType = TYPES.QUERY) {
             }
         };
 
-        res.set('X-SQLAPI-Log', stringifyForLogs(logObj));
+        res.set('X-SQLAPI-Log', stringifyForLogs(logObj).substring(0, HEADER_HARD_LIMIT));
```
oleurud: Is it not enough to limit the queries size? Doing something similar to what you are doing in https://github.com/CartoDB/CartoDB-SQL-API/pull/608/files#diff-9f7a520c48a9160af7adc1bd2706e3c3R50 

If the hard limit applies, we will write a "wrong" JSON and it won't be well parsed in Kibana.

I am not saying to change it, only a heads up. If you think it is not a problem, go-ahead

oleurud: I mean, it is better to lose a request log than returns a failing request

Algunenano: No, the queries are logged as "0" : "Select ...", so even if you write 0 characters per query, you still have several added (`"queryNumber" : ""`). You can see an example in [this test](https://github.com/CartoDB/CartoDB-SQL-API/pull/608/files#diff-3f2e57940ec25ca3254559bfbe66cc0dR183) where sending 1000 queries will go well above the 1024 limit (and it that case even above the 4096, so the JSON will be broken).

It's an extreme case but I'd rather have a header limited in size than a good log in Kibana; the other option is to drop that header its size is too big.

oleurud: :+1:

PR #618, opened 2019-09-25
Add Rate limit headers to CORS

### LINKED ISSUE #617: Add Rate limiting headers to CORS
Right now, it's impossible to programmatically read the status of the limits using JS on a browser, we should add all the rate limiting related headers to `Access-Control-Allow-Headers`

rjimenezda:
Fix #617 

It seems like nobody else is calling cors() so I took the liberty to change the function signature to accept an array instead of a string, for readability.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -22,6 +22,7 @@ Announcements:
 * Middlewarify query controller.
 * Set a hard limit on the size of the X-SQLAPI-Log header.
 * Update cartodb-psql to 0.14.0 and use the timeout parameter for pg.query.
+* Expose rate limit headers for CORS requests.
 
 ## 3.0.0
 Released 2019-02-22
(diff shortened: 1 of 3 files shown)


VictorVelarde:
Good 👦 scout @rjimenezda

rjimenezda:
On acceptance I realized this is actually wrong, the header to be able to read from the browser is `Access-Control-Expose-Headers` :man_facepalming:

rjimenezda:
@oleurud @dgaubert This would be a good time to think if we should expose more headers.

What about X-Cache? Should we let JS users know whether the request is a cache hit / miss? It's not indicated by the response code afaik.

oleurud:
I think we can think about other headers unless @dgaubert has it very clear

rjimenezda:
I've tested it (it's on staging cloud), should we merge & release then?

dgaubert:
Please, update NEWS before merging. Thanks

dgaubert:
After merging, you should deploy `master` in both: `staging` and `production`. No need for releasing though.

PR #619, opened 2019-10-01
Implement router hierarchy to be able to attach middlewares to different routers

dgaubert:
1 of 3 PRs:
- https://github.com/CartoDB/CartoDB-SQL-API/pull/619
- https://github.com/CartoDB/CartoDB-SQL-API/pull/620
- https://github.com/CartoDB/CartoDB-SQL-API/pull/622

In this first PR, I had to refactor a bit by extracting some functionality with middlewares. Besides, I applied some ES6/7/8 goodies.

For CRers, review the following commits:
 - https://github.com/CartoDB/CartoDB-SQL-API/commit/158f28c8e748a42cddd12934c0acd014ad1a6bc5
 - https://github.com/CartoDB/CartoDB-SQL-API/commit/6c775c1857d554d92ba7d33baa1e5ae41b4593c2
- https://github.com/CartoDB/CartoDB-SQL-API/pull/619/commits/07cf020923bdf54cc87f32d6815a6bc25ed89378


app/controllers/health_check_controller.js
```diff
@@ -1,35 +1,40 @@
...
+        }
+
+        const startTime = Date.now();
+
+        healthCheckBackend.check((err) => {
+            const ok = !err;
+            const response = {
                 enabled: true,
-                ok: ok,
+                ok,
```
Algunenano: I don't understand this change. What are the values of response? Does response.ok exists any more?

app/middlewares/cors.js
```diff
@@ -21,6 +21,10 @@ module.exports = function cors(extraHeaders = []) {
         res.header('Access-Control-Allow-Headers', headers.join(', '));
         res.header('Access-Control-Expose-Headers', exposedHeaders.join(', '));
 
+        if (req.method === 'OPTIONS') {
```
Algunenano: You are introducing a different feature here, which is not all that nice to do and doesn't have a NEWS item (in this PR at least).

app/server.js
```diff
@@ -1,69 +1,33 @@
...
-var UserLimitsService = require('./services/user_limits');
-var BatchLogger = require('../batch/batch-logger');
-var JobPublisher = require('../batch/pubsub/job-publisher');
-var JobQueue = require('../batch/job_queue');
-var JobBackend = require('../batch/job_backend');
-var JobCanceller = require('../batch/job_canceller');
-var JobService = require('../batch/job_service');
+const express = require('express');
+const fs = require('fs');
+const mkdirp = require('mkdirp');
```
Algunenano: mkdirp seems abandoned. Do you think we should consider changing it?

Algunenano: In a different PR that is

app/controllers/health_check_controller.js
dgaubert: ES6 property shorthand: http://es6-features.org/#PropertyShorthand.

app/middlewares/cors.js
dgaubert: If I'm not wrong, it should behave in the same way. Having a generic middleware (which is bound for all endpoints) and responds when the method is `options` should be equivalent to defining a `app.options('*', ...)` where basically means for any endpoint being accessed with `options` respond immediately (the cors middleware already took place). 

In any case, I'll double-check it.

app/server.js
dgaubert: 👌

PR #633, opened 2019-11-19
Fix COPY Throttling bug

jgoizueta:
This fixes one of the observed COPY problems: the unexpected "Connection closed by server: input data too slow" errors.

Instead of checking that in the latest N seconds the minimum read bytes was less than M in each one them we were checking that in any different N seconds had less than M bytes.

I've been unable to write a test to demonstrate this case so far.

### DIFF
--- a/lib/services/throttler-stream.js
+++ b/lib/services/throttler-stream.js
@@ -8,34 +8,19 @@ module.exports = class Throttler extends Transform {
 
         this.pgstream = pgstream;
 
-        this.sampleLength = global.settings.copy_from_maximum_slow_input_speed_interval || 15;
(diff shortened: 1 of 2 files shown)


Algunenano:
Can't we get rid of `bytesPerSecondHistory`? We could simply have a counter and increase it if nothing is received in that second or reset it if we did, and if the counter reaches the threshold we killed the request.

dgaubert:
> I've been unable to write a test to demonstrate this case so far.

Now I remember it wasn't easy to implement the test that works the other way around. In any case, we have several tests that validate the COPY FROM/TO are working. See:

- https://github.com/CartoDB/CartoDB-SQL-API/blob/master/test/acceptance/copy-endpoints-test.js  

So, I'd try to not spend much time to implement that test.

PR #634, opened 2019-11-20
Simplify throttling logic

jgoizueta:
This introduces a couple of changes in the parameter interpretation:
* `copy_from_minimum_input_speed` is now the total number of bytes in the interval (so we could multiply the currently configured values by the interval to preserve min. intended throughput.
* the minimum speed is now inclusive (only smaller values will cause an exception).

@Algunenano will this satisfy you? 😛 
@dgaubert do you think this looks safe enough?

### DIFF
--- a/lib/services/throttler-stream.js
+++ b/lib/services/throttler-stream.js
@@ -8,15 +8,15 @@ module.exports = class Throttler extends Transform {
 
         this.pgstream = pgstream;
 
-        this.sampleLength = global.settings.copy_from_maximum_slow_input_speed_interval || 15;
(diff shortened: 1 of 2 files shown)


Algunenano:
> @Algunenano will this satisfy you? stuck_out_tongue

LGTM but it's confusing, it's saying `this.minimunBytesPerSecondThreshold` but the threshold isn't per second, but per `this.sampleLength`.

dgaubert:
Resolve conflict before proceeding, please move on. We are refactoring/optimizing a 10 lines algorithm but we still don't know if we solve the issue or not.

Algunenano:
Can you add a line in NEWS about this change, please?

PR #640, opened 2020-02-12
Allow metrics custom headers in cors

esloho:
As requested in https://app.clubhouse.io/cartoteam/story/58905/implement-client-metrics-in-kepler-sql#activity-58912 we must add the new custom headers for metrics to the allowed headers list for cors.

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -3,6 +3,8 @@
 ## 5.0.2
 Released 2020-mm-dd
 
+- Updated allowed list of custom headers with the ones for metrics
+
 
 ## 5.0.1
 Released 2020-01-27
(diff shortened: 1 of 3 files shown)


dgaubert:
Please update NEWS.md before merging

esloho:
I've seen is the common format but it should match any case format. Even the [documentation](Carto-Source-Lib, Carto-Source-Context, Carto-Source-Context-Id) uses "origin" in the examples for the "Origin" header so I'll say it's ok.

VictorVelarde:
We will need these headers for `api/v2/sql` but also for `api/v2/sql/copyfrom` and `api/v2/sql/copyto`endpoints. Will they work for both in the same way?

esloho:
These headers are sent by our cors protocol, so they will be accepted for to all the request :)

PR #641, opened 2020-02-17
Rename accepted headers for metrics events

esloho:
After reviewing current events we agreed to rename the headers so they are closer to their corresponding event attributes.

\cc @VictorVelarde


esloho:
It seems npm is having problems today, so the tests won't pass. Let's wait until it gets stable again :pray:

VictorVelarde:
I'm ok with the new names. Feel free to move on and please ping me when they're activated on prod, so I can adapt & test the client at toolkit

VictorVelarde:
It looks like it is stable now! 🚀

PR #642, opened 2020-02-18
Add pubsub metrics

esloho:
Closes [ch58698] [ch58699] https://app.clubhouse.io/cartoteam/story/58699/add-pubsub-middleware-to-sql-api

This PR adds a new middleware for sending metrics as pubsub events. This new middleware should act as follows:
- If some headers are sent in the request, it should send an event to pubsub with the information from those headers plus a few additional fields (user_id, a timestamp, event version, response status...)
- This middleware should be asynchronous and not wait for pubsub to finish.

Ideally we wanted to have the result of the pubsub interaction (event sent / error) in kibana like we do in rollbar for cartodb and central. However, after talking with @oleurud he realized that it's not possible with the current implementation of the log middleware:
- the log middleware writes the content to be logged in the response headers
- those are later detected and sent to the corresponding logs used for feeding kibana. 
- this middleware has to be executed before the response is sent. 
- ---> we can't use the log middleware in the pubsub one unless we then wait for pubsub to finish so we can write the results in the logs before sending the response.

The alternative is to write the regular console logs (just in case we desperately need to check the presence of errors) and delegate the monitoring to the later steps (pubsub subscription, criterion, BQ...) /cc @ilbambino 

**Acceptance:**
- Regular behavior should not be altered (navigate and perform some queries on staging to check everything goes well). No events should be generated since no request would add the headers now.
- Run some curl/postman requests with the new headers and check new events are created in the staging project in google cloud (ping me if you don't have access).


test/acceptance/pubsub-metrics-test.js
```diff
@@ -0,0 +1,91 @@
+'use strict';
```
dgaubert: As you are mocking `PubSubMetricsService` this suite test cannot be placed as an acceptance test. Either you move these tests to the `integration` folder or avoid mocks.

test/unit/pubsub-metrics-test.js
```diff
@@ -0,0 +1,86 @@
+'use strict';
```
dgaubert: Just thinking out loud... I've used `sinon` a lot in the past, and it's a great mocking library but I'm getting older 👴 and I'm not sure it pays off, I don't see much benefit of adding a new library just to check if we are calling a method or not. Sometimes it's better to not cover some functionally than mocking everything and test it manually.

Let's keep `sinon` for the moment, but it'd be great if you find a better way to test it without mocking he world.

PS: If I recall correctly, we used to override method to check if they were called or not. Somenthing like this: 

```js
// setup
originalMethod = MyObject.prototype.myMethod;
let called = false;
MyObject.prototype.myMethod = function () { called = true };
// test
myObject.myMethod();
// tear down
MyObject.prototype.myMethod = originalMethod ;
```
But I don't like it, I prefer having `sinon` for that.

esloho:
Thanks a lot @dgaubert !! I love picky reviews, it's when I learn the most. I'll answer some of the thing inline :)

lib/api/sql/copy-controller.js
```diff
@@ -87,7 +87,10 @@ function handleCopyTo (logger) {
 
                     return next(err);
                 })
-                .on('end', () => metrics.end(streamCopy.getRowCount()))
+                .on('end', () => {
+                    metrics.end(streamCopy.getRowCount());
+                    next();
```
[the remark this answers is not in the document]
esloho: My IDE has been complaining about this a lot, but I thought it was _common practice_ in the project. So I'm changing the ones in the new middleware as well :D

lib/services/pubsub-metrics.js
```diff
@@ -0,0 +1,81 @@
...
+
+    getTopic () {
+        if (!this.topic) {
+            this.topic = this.pubsub.topic(this.topicName);
+        }
+
+        return this.topic;
+    }
+
+    isDisabled () {
```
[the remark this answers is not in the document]
esloho: I agree. The thing about checking if the pubsub client or the topic are _alive_ was for preventing wasting time in case of errors. But I find your proposal cleaner and we can explicitly check for errors elsewhere.

lib/services/pubsub-metrics.js
```diff
@@ -0,0 +1,81 @@
...
+        return this.pubsub === undefined || this.topic === undefined;
+    }
+
+    publish (event, attributes) {
```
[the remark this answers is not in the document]
esloho: It should be private, yes, but I need access to mock it... in the current implementation. I'll change it if I can change the tests.

lib/api/middlewares/pubsub-metrics.js
```diff
@@ -0,0 +1,43 @@
...
+    if (pubSubMetrics.isDisabled()) {
+        return function pubSubMetricsDisabledMiddleware (req, res, next) { next(); };
+    }
+
+    return function pubSubMetricsMiddleware (req, res, next) {
+        const [event, attributes] = getEventAttributes(req, res);
+        if (!event) {
+            return next();
+        }
+        pubSubMetrics.sendEvent(res.locals.user, event, attributes, next);
```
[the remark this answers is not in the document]
esloho: There is a potential issue with using the logger and I think this may not work. I'll find you later if you have the time so we can take a look :)

test/unit/pubsub-metrics-test.js
esloho: Honesty it's been a pain in the... face to get these tests working. I don't want to be doing actual calls to the Google PubSub service and that's why I thought of sinon. I don't like to do mocks like this neither :( I'm going to think about it a little. We could test just the isolated logic so we can use a fake pubsub without mocking everything... ONTOH the inegration of this middleware with the other ones is somehow tricky and I think we should cover it... don't know xD

dgaubert: I guess we have this middleware disabled for the whole test suit but this one, don't we?. In Maps API we use `nock` to intercept the calls to third party services and respond with fake values. But still, I'm not sure if `nock` pays off either.

I guess having good unit tests and decoupling `pub/sub` as much as possible is more than enough. Then making sure it's working in both staging and production as acceptance test of the PR.

lib/services/pubsub-metrics.js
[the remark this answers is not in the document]
dgaubert: make it private by using `_` as prefix: `_publish`

package.json
```diff
@@ -68,7 +69,8 @@
     "nyc": "^15.0.0",
     "shapefile": "0.3.0",
     "sqlite3": "^4.0.6",
-    "zipfile": "0.5.12"
+    "zipfile": "0.5.12",
+    "sinon": "^9.0.0"
```
oleurud: I would install it as dev dependency

lib/api/api-router.js
```diff
@@ -62,6 +65,7 @@ module.exports = class ApiRouter {
             paths.forEach(path => app.use(path, apiRouter));
 
             apiRouter.use(error());
+            apiRouter.use(pubSubMetrics(this.pubSubMetricsService));
```
oleurud: Great dependency injection :)

lib/api/sql/copy-controller.js
```diff
@@ -160,6 +163,8 @@ function handleCopyFrom (logger) {
                         time,
                         total_rows: rows
                     });
+
+                    next();
```
oleurud: I always prefer to use `return next()` to be as clear as possible. Same in `errorHandlerMiddleware` function

lib/services/pubsub-metrics.js
```diff
@@ -0,0 +1,66 @@
...
+
+        return this.pubsub.topic(topicName);
+    }
+
+    sendEvent (event, attributes) {
+        if (!this.enabled) {
+            return;
+        }
+
+        const data = Buffer.from(event);
```
oleurud: I don't know why the Google library wants to send a buffer, but if it is needed, I will validate the `event` data type and length since it is an open field that comes from a user header and it could be anything.

package.json
esloho: It is installed as a devDependency though, right?

lib/api/sql/copy-controller.js
esloho: Totally, I just missed it. Thanks!

lib/services/pubsub-metrics.js
esloho: The type will always be string since it's a header, right? Although it doesn't hurt the check :)

About the size... do you mean like checking it isn't longer than X characters?

esloho: In any case we should check that in the middleware, since it's where we "prepare" the data from the headers. Adding it now.

esloho:
> I don't know if it would help much, but I would probably [trim](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/trim) it also

We're allowing blank spaces in central and cartodb events, like for the plan names i.e. That's why I thought we shouldn't but maybe for name, source and groupId we don't want to allow that...?

ilbambino:
but `trim` only removes the white space at the beginning or end, not in the middle. So it should be fine

esloho:
True xDDD

package.json
oleurud: :man_facepalming: right

PR #643, opened 2020-02-20
Do not yield to the next middleware before handleQuery is done

dgaubert:
This PR allows appending middlewares after `handleQuery` middleware in `QueryController`.  Before that, it was yielding before streaming the results of the query into the `Response` and could end up writing in a `Response` which `Express` has actually closed.


dgaubert:
Tested and working like a charm!

Issue #628, opened 2019-10-18
Return 'Retry-after' header when 'Carto-Rate-Limit-Remaining' is 0


rjimenezda:
The header `Retry-After` is extremely useful, since it lets users take advantage of the 'Leaky-bucket' type of limits we have. However, in order to get this header, you must trigger a 429 error.

In practice, a client of the SQL API is going to look at two headers:

- Carto-Rate-Limit-Remaining, to know if they have any remaining requests
- Retry-After, to set up a timer to start firing requests.

As the API stands right now, when you get `Carto-Rate-Limit-Remaining: 0`, you have to make another request (which you know is going to fail), to trigger the error, get the `Retry-After` and know exactly how much you have to wait.

What would be ideal is to return the `Retry-After` header when `Carto-Rate-Limit-Remaining` is 0, as well as on 429 errors.

PR #648, opened 2020-04-06
Return 'Retry-after' header when 'Carto-Rate-Limit-Remaining' is 0.

mondeja:
As was reported in #628, is useful return `Retry-after` header when `Carto-Rate-Limit-Remaining` is 0 to avoid force 429 error in order to get seconds remaining until the next permitted call to the API for an user with rate limits enabled.

### DIFF
--- a/lib/api/middlewares/rate-limit.js
+++ b/lib/api/middlewares/rate-limit.js
@@ -31,8 +31,8 @@ function rateLimit (userLimits, endpointGroup = null) {
                 'Carto-Rate-Limit-Remaining': remaining,
                 'Carto-Rate-Limit-Reset': reset
             });
-            
-            if (remaining == 0 && !isBlocked) {
+
+            if (remaining === 0 && !isBlocked) {
                 // retry is floor rounded in seconds by redis-cell
                 res.set('Retry-After', retry + 2);
             }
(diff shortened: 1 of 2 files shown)


dgaubert:
Hi @mondeja! 

Thanks for contributing. As per https://github.com/CartoDB/CartoDB-SQL-API#contributing, external contributions must sign a CLA. Let me contact with CARTO's legal department to know how to proceed.

Apart from the legal stuff, the linter is failing because of:

```
/srv/lib/api/middlewares/rate-limit.js
  34:1   error  Trailing spaces not allowed          no-trailing-spaces
  35:27  error  Expected '===' and instead saw '=='  eqeqeq
✖ 2 problems (2 errors, 0 warnings)
  1 error and 0 warnings potentially fixable with the `--fix` option.
```

cc/ @oleurud as he's the :brain: behind rate-limiting.

lib/api/middlewares/rate-limit.js
```diff
@@ -32,8 +32,12 @@ function rateLimit (userLimits, endpointGroup = null) {
                 'Carto-Rate-Limit-Reset': reset
             });
 
-            if (isBlocked) {
+            if (remaining === 0 && !isBlocked) {
                 // retry is floor rounded in seconds by redis-cell
+                res.set('Retry-After', retry + 2);
```
oleurud, line 37: `retry` is `-1` if  `isBlocked` is false, so, `retry + 2` will always be 1. That will work perfectly if the rate limit is "per second" (being strict, if we are going to give a new token in less than a second), but it won't if the rate limit is "per minute" (being strict, if we are going to give a new token in more than a second). Let me think about it

PR #660, opened 2020-06-22
respond with user-id and client in the headers

ilbambino:
added headers to the test-client callback to be able to check them

### DIFF
--- a/lib/api/middlewares/user.js
+++ b/lib/api/middlewares/user.js
@@ -15,7 +15,7 @@ module.exports = function user (metadataBackend) {
                 error.subtype = 'user-not-found';
                 error.http_status = 404;
                 error.message = errorUserNotFoundMessageTemplate(res.locals.user);
-                next(error);
+                return next(error);
             }
 
             res.locals.userId = userId;
--- a/lib/api/middlewares/user.js
+++ b/lib/api/middlewares/user.js
@@ -19,6 +19,7 @@ module.exports = function user (metadataBackend) {
             }
 
             res.locals.userId = userId;
+            res.set('Carto-User-Id', `${userId}`);
             return next();
         });
     };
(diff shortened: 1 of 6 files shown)


lib/api/sql/query-controller.js
```diff
@@ -55,6 +56,7 @@ module.exports = class QueryController {
                 lastModified(),
                 formatter(),
                 content(),
+                clientHeader(),
```
dgaubert, line 59: I guess we also want it for Batch Queries and Copy From/to endpoints, don't we?

If so, I'd use this middleware in `sql-router`, which is the place where we use all common middlewares for the public API.

test/acceptance/client-headers-test.js
```diff
@@ -0,0 +1,42 @@
...
+
+        this.testClient.getResult(publicSQL, (err, result, headers) => {
+            assert.ifError(err);
+            assert.strictEqual(result.length, 6);
+            assert.strictEqual(headers['carto-client'], undefined);
+            done();
+        });
+    });
+
+    it('should not get the user id in the response header', function (done) {
```
dgaubert, line 32: the description doesn't correspond with the test implementation, does it?

lib/api/sql/query-controller.js
dgaubert: Also, the tests are failing because of the order you placed the middleware. It should run before the response has sent the first byte to the client.

test/acceptance/client-headers-test.js
ilbambino: 🤕 upps. Some  `not` that shouldn't be there.

lib/api/sql/query-controller.js
ilbambino: wasn't really the order but a missing return in case of error. But it should be fixed now :)

ilbambino:
[ch87061]

lib/api/middlewares/user.js
```diff
@@ -15,10 +15,11 @@ module.exports = function user (metadataBackend) {
                 error.subtype = 'user-not-found';
                 error.http_status = 404;
                 error.message = errorUserNotFoundMessageTemplate(res.locals.user);
-                next(error);
+                return next(error);
```
dgaubert, line 18: Good catch! :)

PR #663, opened 2020-06-30
Log system revamp

dgaubert:
Breaking changes:
- Log system revamp:
  - Logs to stdout, disabled while testing
  - Use header `X-Request-Id`, or create a new `uuid` when no present, to identyfy log entries
  - Be able to set log level from env variable `LOG_LEVEL`, useful while testing: `LOG_LEVEL=info npm test`; even more human-readable: `LOG_LEVEL=info npm t | ./node_modules/.bin/pino-pretty`
  - Be able to reduce the footprint in the final log file depending on the environment
  - Use one logger for every service: Queries, Batch Queries (Jobs), and Data Ingestion (CopyTo/CopyFrom)
  - Stop using headers such as: `X-SQL-API-Log`, `X-SQL-API-Profiler`, and `X-SQL-API-Errors` as a way to log info.
  - Be able to tag requests with labels as an easier way to provide business metrics
  - Metro: Add log-collector utility (`metro`), it will be moved to its own repository. Attaching it here fro development purposes. Try it with the following command `LOG_LEVEL=info npm t | node metro`
  - Metro: Creates `metrics-collector.js` a stream to update Prometheus' counters and histograms and exposes them via Express' app (`:9145/metrics`). Use the ones defined in `grok_exporter`

Announcements:
- The profiler is always set. No need to check its existence anymore
- Unify profiler usage for every endpoint

Bug fixes:
- Avoid hung requests while fetching user identifier

cc @ilbambino


app.js
```diff
@@ -81,89 +53,64 @@ if (global.settings.statsd) {
...
-                });
+                process.exit(code);
             });
         });
-    };
+    });
+});
+
+function addHandlers (listener, killTimeout) {
+    // FIXME: minimize the number of 'uncaughtException' before uncomment the following line
```
oleurud, line 100: And what is the plan?

lib/api/sql/query-controller.js
```diff
@@ -1,14 +1,13 @@
...
 const bodyParser = require('../middlewares/body-parser');
-const { initializeProfiler } = require('../middlewares/profiler');
 const user = require('../middlewares/user');
 const rateLimits = require('../middlewares/rate-limit');
 const authorization = require('../middlewares/authorization');
 const connectionParams = require('../middlewares/connection-params');
 const timeoutLimits = require('../middlewares/timeout-limits');
 const params = require('../middlewares/params');
-const log = require('../middlewares/log');
+const log = require('../middlewares/log-query');
```
oleurud, line 10: log => logQuery?

lib/batch/models/job-fallback.js
```diff
@@ -205,54 +205,39 @@ JobFallback.prototype.getLastFinishedStatus = function () {
...
-        }
-
-        logger.info(logEntry, tag);
-    }
-
-    return true;
+JobFallback.prototype.toJSON = function () {
+    const queries = this.data.query.query;
+
+    return {
```
oleurud, line 211: :muscle:

app.js
dgaubert: I'm keeping this comment from the old implementation of this. So I guess the plan is to FIX IT as soon as possible but it's out of the scope of this PR.

lib/api/sql/query-controller.js
dgaubert: it was used in both Job and Query controllers. Now as jobs know how to outputs its own information (`.toJSON()`) this middleware is no longer needed for Job controller. The middleware rename is just to reflect it.

PR #669, opened 2020-10-08
Disable streaming mode by redis field "streaming_mode"

manmorjim:
:warning: This [PR](https://github.com/CartoDB/node-cartodb-redis/pull/34) has to be merged and deployed before merging this.

### Changes

Reading the field `streaming_mode` from db parameters of redis which indicates the operation mode for the SQL API. If `true` (default) then the SQL API streams the results (default behaviour). If this field is set to `false` the the SQL API will return all the results at once.

More info in CH: https://app.clubhouse.io/cartoteam/story/106733/turn-sql-api-transactional-and-improve-the-error-handling

### DIFF
--- a/lib/api/middlewares/formatter.js
+++ b/lib/api/middlewares/formatter.js
@@ -4,10 +4,20 @@ const formats = require('../../models/formats');
 
 module.exports = function formatter () {
     return function formatterMiddleware (req, res, next) {
-        const { format } = res.locals.params;
+        const { params: { format },  userDbParams: { streamingmode } } = res.locals;
+
+        /* parses streamingmode to boolean. This is:
+          streamingmode = undefined --> true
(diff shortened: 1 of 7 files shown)


dgaubert:
What happens when the user that has the `streaming mode` disabled sends a query for a table of 10GB?

manmorjim:
> What happens when the user that has the `streaming mode` disabled sends a query for a table of 10GB?

That's a good point! We should establish a quota limit for those cases where the `streaming_mode` is disabled. This limit could be defined in the config such as `db_max_row_size`.

dgaubert:
what happens when the available memory is less than the threshold or even the memory needed to handle that query?.

thedae:
Closing since these changes are not going to be used

PR #670, opened 2020-10-22
Fix row_too_large error handling

thedae:
The `return console.error` was preventing the stream to be correctly closed through `handleQueryEnd`

Removing the `return` fixes the weird behaviour and now the errors are consistent


dgaubert (approved):
Sorry for the delay!

PR #676, opened 2021-03-01
add dev-env tests

alberhander:
- Adding dev-env tests on pr builds.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -67,7 +67,7 @@ jobs:
 
   dev-env-tests:
     runs-on: ubuntu-18.04
-    timeout-minutes: 10
+    timeout-minutes: 15
     needs: build-and-test
 
     steps:
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -65,6 +65,7 @@ jobs:
     - uses: actions/checkout@v2
       with:
         repository: 'cartodb/docker-dev-env'
+        path: 'docker-dev-env'
         token: ${{ secrets.CARTOFANTE_PERSONAL_TOKEN }}
 
     - name: Check docker-compose syntax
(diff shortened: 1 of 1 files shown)


.github/workflows/main.yml
```diff
@@ -60,6 +61,51 @@ jobs:
...
+        docker stop postgres redis
+
+    # dev-env tests
+    - uses: actions/checkout@v2
+      with:
+        repository: 'cartodb/docker-dev-env'
+        path: 'docker-dev-env'
+        token: ${{ secrets.CARTOFANTE_PERSONAL_TOKEN }}
+
+    - name: Check docker-compose syntax
```
ilbambino, line 75: Do you think this step adds value? In `dev-env` where the compose file changes I agree, but here?

ilbambino (commented):
would it make sense to do it as another job? 
if it fails it would be easier to know why. 
I don't know if we could use the image from one job and reuse it in others (I haven't read the docs if possible).
Also if we add later (soon) tests for onprem they could run in parallel and if they are separated jobs the failures would be clearer

.github/workflows/main.yml
alberhander: I just left it for debugging purposes, because the docker-compose file lives in a different repository.

alberhander:
> would it make sense to do it as another job?
> if it fails it would be easier to know why.
> Once the image is built it could be [cached](https://docs.github.com/en/actions/guides/caching-dependencies-to-speed-up-workflows) and reused in the other jobs. (add a dep using `needs: build-and-test`)
> Also if we add later (soon) tests for onprem they could run in parallel and if they are separated jobs the failures would be clearer

I didn't add a different job because most of the repos are in Cloud Build and there's no such option there, but it makes sense doing it here, it will be easier to troubleshoot any failure. I'll make them work (I'm missing gcr permissions) and I'll run them in a separate job!

PR #678, opened 2021-03-09
build onprem image

alberhander:
- Related to: https://app.clubhouse.io/cartoteam/story/139906/include-licensing-system
- Depends on: https://github.com/CartoDB/CartoDB-SQL-API-private/pull/1 (bump `private` after merge).

### DIFF
--- a/lib/api/middlewares/error.js
+++ b/lib/api/middlewares/error.js
@@ -2,13 +2,13 @@
 
 const errorHandlerFactory = require('../../services/error-handler-factory');
 
-module.exports = function error () {
+module.exports = function error ({ logger }) {
     return function errorMiddleware (err, req, res, next) {
-        const { logger } = res.locals;
         const errorHandler = errorHandlerFactory(err);
         const errorResponse = errorHandler.getResponse();
+        const errorLogger = res.locals.logger || logger;
 
-        logger.error({ exception: err }, 'Error while handling the request');
+        errorLogger.error({ exception: err }, 'Error while handling the request');
 
         // Force inline content disposition
         res.header('Content-Disposition', 'inline');
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -125,9 +125,9 @@ jobs:
           docker run --rm --name tavern-tester \
             --env CARTO_USER=carto  \
             --env CARTO_DOMAIN=localhost.lan \
-            --network carto_default \
+            --network host \
             --env-file \
-            .env gcr.io/cartodb-on-gcp-main-artifacts/tavern-tester \
+            .env gcr.io/cartodb-on-gcp-main-artifacts/tavern-tester:latest \
             all dev
 
     - name: Push image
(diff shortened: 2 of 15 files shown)


ilbambino (approved):

alberhander:
Thanks a lot @thedae !! 

@dgaubert I'm trying to add (in `config/environments/config.js` the middleware configuration conditionally depending on a env var only used in onprem, but it doesn't seem to work. In case you can suggest me how to do that:
```
// load onprem licensing middleware
if (process.env.CARTO_SQL_ONPREM_MIDDLEWARE) {
  module.exports.routes.api[0].middlewares = require('/usr/src/app/onprem.js').middlewares
} else {
  module.exports.routes.api[0].middlewares = [
      function noop () {
          return function noopMiddleware (req, res, next) {
              next();
          }
      }
  ]
}
```

lib/api/middlewares/error.js
```diff
@@ -2,13 +2,13 @@
 
 const errorHandlerFactory = require('../../services/error-handler-factory');
 
-module.exports = function error () {
+module.exports = function error ({ logger }) {
     return function errorMiddleware (err, req, res, next) {
```
dgaubert, line 6: With this change, the logger is not going to log with the request Id (see: https://github.com/CartoDB/CartoDB-SQL-API/blob/master/lib/api/middlewares/logger.js#L8), so we are going to lose the ability to correlate the error with the user's request.

dgaubert: In any case, I understand the reason for the fallback of using the main logger when the child logger is not available. It was just a "heads up".

thedae: Thanks for pointing this out, it helps :)

I didn't want to modify (and probably break) anything else, but another solution would be to create a new middleware to actually instantiate the child logger into `res.locals.logger` and remove that from the `initLogger` one. Then we could use this new middleware across the whole application, not only specific endpoints

What do you think?

thedae:
> Thanks a lot @thedae !!
> 
> @dgaubert I'm trying to add (at the bottom of `config/environments/config.js` the middleware configuration conditionally depending on a env var only used in onprem, but it doesn't seem to work. In case you can suggest me how to do that:
> 
> ```
> // load onprem licensing middleware
> if (process.env.CARTO_SQL_ONPREM_MIDDLEWARE) {
>   module.exports.routes.api[0].middlewares = require('/usr/src/app/onprem.js').middlewares
> } else {
>   module.exports.routes.api[0].middlewares = [
>       function noop () {
>           return function noopMiddleware (req, res, next) {
>               next();
>           }
>       }
>   ]
> }
> ```

@alberhander let me try this out, I'll get back to you with a proposal

lib/api/middlewares/error.js
dgaubert: Fair enough, it's fine as it is.

ilbambino:
you are right with the idea that we are building it as a separate image. But what we wanted is to have one single config file and be able to inject the middleware only in the onprem case. 
So the idea was to set that env var in the build phase of the image, so it is "baked in" but without to maintain two config files. Because as it is an array with source code, injecting it through an env var is not so easy/clean. But if you have better ideas, we are all ears 👂

thedae:
I have a proposal for this scenario: we could pass the prepended middlewares in an ENV var as a comma-separated ordered list of paths to those middlewares, for example:
```
middlewares: process.env.CUSTOM_MIDDLEWARES
```
With `CUSTOM_MIDDLEWARES=/usr/src/app/onprem.js,/usr/src/app/another-one.js,/usr/src/app/cool-stuff.js`

This would require a small change in the piece of code where we evaluate this config entry, fairly easy to achieve.

What do you think @dgaubert @ilbambino @alberhander  ?

ilbambino:
I like the idea 👍

dgaubert:
Go for it! In previous on-prem releases, we couldn't use env variables. If the new process to build on-prem allows them, your approach is wide better.

alberhander:
We can ignore the `dev-env` tests failing, nothing to do with the sql-api

alberhander:
Already tested the image in on-prem, and it works! :muscle:

alberhander:
Tested again, and working as expected :+1:

PR #679, opened 2021-03-23
[ch140184] Update documentation about fallback templates

Shylpx:
### Resources

- [Clubhouse story](https://app.clubhouse.io/cartoteam/story/140184/gloval-seven-admin-batch-query-fallback-templates-not-being-replaced)

### Context

- Fallback templates were implemented only at the query level [here](https://github.com/CartoDB/CartoDB-SQL-API/blob/master/lib/batch/models/query/fallback.js#L30-L59). Fallbacks are also implemented at the job level [here](https://github.com/CartoDB/CartoDB-SQL-API/blob/master/lib/batch/models/query/main-fallback.js#L31-L49), but without any template. 
- However, the example in the documentation is using the fallback template at the job level, storing the information about the error when the table name in the query is invalid, and this can only be performed at the query level (even if there were templates also for the job, because the `error_message` is not propagated).

### Changes

- Updating the documentation, by using the fallback templates at the query level.
- Removing the message about new templates, since I guess that is not true at this point :eyes:.

### DIFF
--- a/docs/guides/03-batch-queries.md
+++ b/docs/guides/03-batch-queries.md
@@ -460,14 +460,12 @@ This is helpful when you want to save error messages into a table:
 curl -X POST -H "Content-Type: application/json" -d '{
   "query": {
     "query": [{
-      "query": "UPDATE wrong_table SET price = '$100.00' WHERE company = 'CARTO'"
-    }],
-    "onerror": "INSERT INTO errors_log (job_id, error_message, date) VALUES ('<%= job_id %>', '<%= error_message %>', NOW())"
+      "query": "UPDATE wrong_table SET price = '$100.00' WHERE company = 'CARTO'",
+      "onerror": "INSERT INTO errors_log (job_id, error_message, date) VALUES ('<%= job_id %>', '<%= error_message %>', NOW())"
+    }]
   }
 }' "https://{username}.carto.com/api/v2/sql/job"
 ```
-
-More templates are coming soon.
 
 ### Fetching Job Results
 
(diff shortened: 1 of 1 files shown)


dgaubert:
IIRC you can define both fallbacks, at query level and job level:
- Main (job) fallback: https://github.com/CartoDB/CartoDB-SQL-API/blob/master/lib/batch/models/query/main-fallback.js
- Query fallback: https://github.com/CartoDB/CartoDB-SQL-API/blob/master/lib/batch/models/query/fallback.js

dgaubert:
Not every job type supports main fallback, see: https://github.com/CartoDB/CartoDB-SQL-API/tree/master/lib/batch#fallback

Shylpx:
@dgaubert You can define both fallbacks, but not templates for both, right? I mean, as you said, you can define fallbacks at query level and at job level ([documentation](https://github.com/CartoDB/CartoDB-SQL-API/blob/0787cc8bb5afa4ba1f41ea2d67c4dd8ffb608b18/docs/guides/03-batch-queries.md#chaining-batch-queries-with-fallbacks)). But AFAIK fallback **templates** are only implemented at [query level](https://github.com/CartoDB/CartoDB-SQL-API/blob/master/lib/batch/models/query/fallback.js#L35), so I'm trying to update only the [documentation for that subsection](https://github.com/CartoDB/CartoDB-SQL-API/blob/0787cc8bb5afa4ba1f41ea2d67c4dd8ffb608b18/docs/guides/03-batch-queries.md#templates).

dgaubert:
Perfect then, nice job.

dgaubert (approved):

PR #680, opened 2021-03-24
[ch139802] Avoid concurrent requests from getting stuck when exporting file formats handled by OGR

Shylpx:
### Resources

- [Clubhouse story](https://app.clubhouse.io/cartoteam/story/139802/phl-inconsistent-errors-when-downloading-datasets-through-sql-api)

### Context

- Information about the issue and the suggested solution in [this comment](https://app.clubhouse.io/cartoteam/story/139802/phl-inconsistent-errors-when-downloading-datasets-through-sql-api#activity-144082).

### Changes

- Clean code about "baking" requests. Now, each request is handling its own response.
- Add `close` event to the input stream, in order to do the cleaning tasks even if the output stream is closed without reading anything (the `end` event is not triggered). For example, when a 304 response is obtained by the client.
- Add a random value to the temporal file name to avoid problems with concurrent requests.

### DIFF
--- a/.github/workflows/main.yml
+++ b/.github/workflows/main.yml
@@ -11,6 +11,7 @@ env:
   ARTIFACTS_PROJECT_ID: cartodb-on-gcp-main-artifacts
   NODE_VERSION: '12.18.3'
   _DOCKER_IMAGE_TO_REPLACE: 'sql-api'
+  COMPOSE_PROJECT_NAME: 'carto'
 
 jobs:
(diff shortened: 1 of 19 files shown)


thedae (approved):
LGTM! Thanks for your thorough analysis!

dgaubert (approved):
Great job! Impressive you were able to follow and modify this part of the code which is hard to understand.

PR #685, opened 2021-04-07
Be able to run tests in development env w/o requiring a different dokerfile image

dgaubert:
cc/ @oleurud


test/index.js
```diff
@@ -1,142 +1,193 @@
...
+const fs = require('fs').promises;
 
 if (!process.env.NODE_ENV) {
     console.error('Please set "NODE_ENV" variable, e.g.: "NODE_ENV=test"');
     process.exit(1);
 }
-
 let configFileName = process.env.NODE_ENV;
-if (process.env.CARTO_SQL_API_ENV_BASED_CONF) {
+if (process.env.CARTO_WINDSHAFT_ENV_BASED_CONF) {
```
thedae, line 12: I think this env var is not correct, shouldn't it be `CARTO_SQL_API_ENV_BASED_CONF` ?

thedae (changes requested):
Just a minor comment

test/index.js
dgaubert: haha! `copy/paste`. Nice catch!

thedae (approved):
:+1:

PR #696, opened 2021-11-22
[sc189175] Use cursors on PG queries

Shylpx:
### Resources

- [Shortcut story](https://app.shortcut.com/cartoteam/story/189175/sql-api-oom-errors-when-exporting-data)

### Changes

- ~Use `export_timeout` in OGR exports instead of the render timeout.~ -> https://github.com/CartoDB/CartoDB-SQL-API/pull/698
- Update PG formats based on streams (`json`, `geojson`, `svg`) , to use cursors.

### DIFF
--- a/lib/api/middlewares/formatter.js
+++ b/lib/api/middlewares/formatter.js
@@ -8,7 +8,6 @@ module.exports = function formatter () {
 
         const FormatClass = formats[format];
         req.formatter = new FormatClass();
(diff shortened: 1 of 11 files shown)


Shylpx:
After increasing servers' memory, the error disappear, so using cursors is not needed anymore: https://app.shortcut.com/cartoteam/story/189175/sql-api-oom-errors-when-exporting-data#activity-198251

PR #698, opened 2021-12-28
[sc189175] Add export timeout for OGR exports

Shylpx:
### Resources

- [Shortcut story](https://app.shortcut.com/cartoteam/story/189175/sql-api-oom-errors-when-exporting-data)

### Changes

- Use `export_timeout` in OGR exports instead of the render timeout.

### DIFF
--- a/lib/api/middlewares/timeout-limits.js
+++ b/lib/api/middlewares/timeout-limits.js
@@ -12,7 +12,8 @@ module.exports = function timeoutLimits (metadataBackend) {
             }
 
             const userLimits = {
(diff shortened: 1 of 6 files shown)


moicalcob (approved):
LGTM!! 🚀

Shylpx:
After increasing servers' memory, the error disappears, so we don't need to use the `export_timeout` anymore for OGR exports: https://app.shortcut.com/cartoteam/story/189175/sql-api-oom-errors-when-exporting-data#activity-198251

